PR #93, opened 2015-01-21
ES 1.4 updates

TwP:
Making sure everything works with ElasticSearch 1.4. Thus far it has only been test changes to support differences in response bodies from `aliases` methods and `mappings` methods.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -89,6 +89,15 @@ def es_version_1_x?
   $client.semantic_version >= '1.0.0'
 end
 
+# Elasticsearch 1.4 changed the response body for interacting with index
+# alaises. If an index does not contain any aliases, then an "alises" key is no
+# longer returned in the resopsne.
+#
+# Reeturns `true` if the response contains an "alises" key.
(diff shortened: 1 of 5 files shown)


test/test_helper.rb
```diff
@@ -89,6 +89,15 @@ def es_version_1_x?
   $client.semantic_version >= '1.0.0'
 end
 
+# Elasticsearch 1.4 changed the response body for interacting with index
+# alaises. If an index does not contain any aliases, then an "alises" key is no
+# longer returned in the resopsne.
+#
+# Reeturns `true` if the response contains an "alises" key.
+def es_version_always_returns_aliases?
+  $client.semantic_version <= '1.4.0'
```
grantr: I'd also add >= 1.4.3

grantr:
:+1:

test/test_helper.rb
TwP: Is it being switched back in the next version of ES?

grantr: Yes: https://github.com/elasticsearch/elasticsearch/pull/9178

PR #96, opened 2015-01-21
Support scroll queries

TwP:
We have support for scan queries, but we do not expose the scroll query. The scroll query allows us to sort results. The Scan query does not allow sorting, and therefore, it is a bit more efficient than a regular scroll query.

These changes allow us to support both via a Scroller object.


TwP:
I would love to get this merged in before the 0.5.0 release. @dewski this will enable you to do sorted scroll queries via the elastomer-client gem. That feature was an oversight on my part.

lib/elastomer/client/docs.rb
```diff
@@ -415,8 +415,37 @@ def bulk( params = {}, &block )
...
+      # query  - The query to scroll as a Hash or a JSON encoded String
+      # opts   - Options Hash
+      #   :index  - the name of the index to search
+      #   :type   - the document type to search
+      #   :scroll - the keep alive time of the scrolling request (5 minutes by default)
+      #   :size   - the number of documents per shard to fetch per scroll
+      #
+      # Examples
+      #
+      #   scan = index.scan('{"query":{"match_all":{}},"sort":{"date":"desc"}}')
```
dewski: The docs are still using scan here.

lib/elastomer/client/index.rb
```diff
@@ -413,8 +413,37 @@ def bulk( params = {}, &block )
...
+      # query  - The query to scroll as a Hash or a JSON encoded String
+      # opts   - Options Hash
+      #   :index  - the name of the index to search
+      #   :type   - the document type to search
+      #   :scroll - the keep alive time of the scrolling request (5 minutes by default)
+      #   :size   - the number of documents per shard to fetch per scroll
+      #
+      # Examples
+      #
+      #   scan = index.scan('{"query":{"match_all":{}},"sort":{"date":"desc"}}')
```
dewski: The docs are still using scan here.

dewski:
Other than the simple doc fixes, looks good to me! Thanks for your quick work on scroll support.

lib/elastomer/client/scroller.rb
```diff
@@ -18,12 +19,27 @@ class Client
...
+      Scroller.new(self, query, opts)
+    end
+
+    #
+    #
+    def start_scroll( opts = {} )
+      opts = opts.merge :action => 'search.start_scroll'
+      response = get '{/index}{/type}/_search', opts
+      response.body
+    end
```
grantr: Why are the separate `start_scroll` and `continue_scroll` methods required?

TwP: When you initiate a scroll, you can specify the index and the document type the query will execute against. The initial call returns the `scroll_id`. To continue scrolling you pass this `scroll_id` to the `/_search/scroll` API endpoint. The index and document type are no longer needed since that information is maintained by ES internally and retrieved via the `scroll_id`.

What is now the `start_scroll` method was originally embedded in the `Scan` class as part of the first `scroll_id` method call. It was moved out into its own method for (a) clarity and (b) because the initial results returned by a `scan` query do _not_ contain any document hits. But the initial call with a regular `scroll` query does include document hits. Take a look at the `Scroller#do_scroll` method for the gory details.

grantr: Good explanation thanks! :+1:

grantr:
:ship:

PR #97, opened 2015-04-03
Response body in the notifications payload

TwP:
I've found a need to access the response body in the notifications payload. I would like to measure the different between the request timing and the search time reported by ES in the response. The difference between the two timing values is the network latency.


grantr:
:+1:

grantr:
Can we get a test or two in https://github.com/github/elastomer-client/blob/master/test/notifications_test.rb?

TwP:
I'll throw in a version bump and cut a new gem release, too.

dewski:
:+1:

PR #98, opened 2015-06-23
Fix the build, and some typoes

GregMefford:


grantr:
Thanks @GregMefford! Sorry we didn't notice the build breaking.

Could you change the ES 1.4.0 check to a test_helper method? Perhaps `supports_update_mapping?`. That way all the version checks are in one place and they tell the reader what they're checking for.

We should probably move them to their own file as well, but that's not necessary to do here.

GregMefford:
I called it `es_version_supports_update_mapping_with__all_disabled?` because it's not that Mapping Updates always fail, just that they fail when the `_all` field is disabled (as is the case in this test).

I thought about just changing the test to not set the `_all` field to disabled, but felt like it's best to test how many production environments will really work, and capture the fact that this edge-case might hit someone in real life just as it has hit us in this test environment.

grantr:
:+1:

PR #99, opened 2015-07-02
Reset http connection on timeouts

### LINKED ISSUE #100: Reopen the connection on timeout
When the `read_timeout` is reached in Elastomer::Client, an exception is raised and the Ruby code aborts the request handling process. The Elasticsearch cluster will continue processing the search request and will eventually send a response - the HTTP connection is still open. Subsequent calls using this same connection can return search results for a previous request that timed out. This scenario is the reason that the [`OpaqueId`](https://github.com/github/elastomer-client/blob/master/lib/elastomer/middleware/opaque_id.rb) middleware exists.

A more proactive solution would be to **not** reuse the connection when a read timeout is reached. Instead, the connection should be discarded and a new connection established. This will prevent subsequent search requests from triggering the OpaqueId error condition.

/cc @grantr @tmm1

tmm1:
Avoids re-using the same http connection after a read timeout, which can lead to reading in the wrong response on the next request (when the elasticsearch server eventually finishes processing the initial request which we timed out reading a response for).

fixes #100

cc @TwP

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -187,6 +187,11 @@ module Elastomer
 
         # wrap Faraday errors with appropriate Elastomer::Client error classes
         rescue Faraday::Error::ClientError => boom
+          # reset connection on timeouts to avoid reading incorrect response on next request
+          if Faraday::Error::TimeoutError === boom
+            @connection = nil
+          end
+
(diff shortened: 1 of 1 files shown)


TwP:
The read timeout theory is a red herring.

PR #101, opened 2015-08-14
Run the snapshot tests only when repo creation is possible

grantr:
On ES 1.6+ creation of FS repositories is impossible unless `repo.path` is set in the config. This causes snapshot and repository tests to fail with a default ES config, which is not great.

Instead of always running snapshot tests, only run them if repo creation succeeds. If repo creation fails, snapshot tests are skipped and a warning is displayed:

```
Could not create a snapshot repo. Snapshot tests will be disabled.
To enable snapshot tests, add a path.repo setting to your elasticsearch.yml file.
```

/cc @TwP

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ sudo: true
 
 before_install:
     - curl -O https://download.elastic.co/elasticsearch/elasticsearch/elasticsearch-1.7.1.deb && sudo dpkg -i --force-confnew elasticsearch-1.7.1.deb
-    - "echo 'repo.path: /tmp/elasticsearch' >> /etc/elasticsearch/elasticsearch.yml"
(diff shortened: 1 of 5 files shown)


TwP:
Thank you for fixing the snapshot tests.

TwP:
One nice enhancement would be to add an explicit call to `skip` in each test. That will provide feedback to the user that some tests are not being run.

```ruby
it 'determines if a snapshot exists' do
  with_tmp_repo do |repo|

    skip("path.repo is not configured") if !run_snapshot_tests?

    snapshot = repo.snapshot(@name)
    assert_equal false, snapshot.exists?
    assert_equal false, snapshot.exist?
    snapshot.create({}, :wait_for_completion => true)
    assert_equal true, snapshot.exist?
  end
end
```

Or something like that. Obviously it will need a little more finesse. Just my $0.02 for now.

PR #102, opened 2015-08-14
Fixing test file patterns

TwP:
Test files located directly under the `test/` folder were not being run. The matching patterns would only find test files located in sub-folders of the `test/` folder.

/cc @grantr

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@ require "bundler/gem_tasks"
 require 'rake/testtask'
 
 Rake::TestTask.new do |t|
-    t.pattern = "test/**/*_test.rb"
+    t.test_files = FileList["test/**/*_test.rb"]
 end
 
 task :default => :test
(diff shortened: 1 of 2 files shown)


grantr:
Good catch @TwP :+1:

TwP:
Looking at the snapshot test skipping, I was playing around with adding `skip "reason"` at the top of a test to see if that actually works. I was confused when nothing happened - I was expecting some sort of output or an exception. That is when I realized that the test file was not being run.

grantr:
@TwP it would be nice to use skips instead of the global `if` at the top of the snapshot tests. If you can make that work that'd be great.

Maybe something like this:

```ruby

def snapshot_test(*args, &block)
  if run_snapshot_tests?
    test(*args, &block)
  else
    test(*args) { skip }
  end
end

snapshot_test "create a snapshot" do
  # ...
end
```

TwP:
@grantr LOL !!!

I just added the same comment to your closed PR. And I just hit the merge button on this one. I'll open another PR for finer grained skipping messages.

PR #103, opened 2015-08-15
Skip individual tests

TwP:
Cheating a little bit here and putting the `skip` statement in the `before` block so that we check if snapshots are possible with the current ES configuration. If not, then each test is skipped and a nice message is displayed to the user telling them how to enable these tests.

/cc @grantr


grantr:
:+1: Thanks @TwP

Issue #106, opened 2015-09-10
Upgrade Scroller to support arbitrary URL parameters


chrismwendt:
Scroller currently only supports a handful of parameters. This prevents things like query string parameters from working on scroll requests: https://github.com/github/elastomer-client/blob/8cd99b8e40658b5986aab8f5e033fca4a58eeebc/lib/elastomer/client/scroller.rb#L133-L137

```ruby
        @index       = opts.fetch(:index, nil)
        @type        = opts.fetch(:type, nil)
        @scroll      = opts.fetch(:scroll, '5m')
        @size        = opts.fetch(:size, 50)
        @search_type = opts.fetch(:search_type, nil)
```

The original `opts` just need to be propagated through the methods.

PR #107, opened 2015-09-10
Propagate original Scroller options

chrismwendt:
This allows for other URL parameters (such as the query string in #106)
to be applied to a scroll operation.

I would like to hear feedback on the approach taken here (using an `@opts` instance variable to bundle the options together) and the code style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.0 (2015-09-11)
+- Support all URL parameters when using `Client.#scroll`
+- BREAKING: Moved some `Scroller` reader methods into `Scroller.opts`
+
 ## 0.5.1 (2015-04-03)
 - Add response body to notification payload
 
--- a/lib/elastomer/client/scroller.rb
+++ b/lib/elastomer/client/scroller.rb
@@ -128,19 +128,22 @@ module Elastomer
       #
       def initialize( client, query, opts = {} )
         @client = client
-        @query  = query
 
-        @index       = opts.fetch(:index, nil)
-        @type        = opts.fetch(:type, nil)
-        @scroll      = opts.fetch(:scroll, '5m')
-        @size        = opts.fetch(:size, 50)
-        @search_type = opts.fetch(:search_type, nil)
+        default_opts = {
+          :index => nil,
+          :type => nil,
+          :scroll => '5m',
+          :size => 50,
+          :body => query,
+        }
+
+        @opts = default_opts.merge opts
 
         @scroll_id = nil
         @offset = 0
       end
 
-      attr_reader :client, :query, :index, :type, :scroll, :size, :search_type, :scroll_id
+      attr_reader :client, :query, :scroll_id
 
       # Iterate over all the search results from the scan query.
       #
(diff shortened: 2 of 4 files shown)


lib/elastomer/client/scroller.rb
```diff
@@ -128,19 +128,22 @@ class Scroller
...
-        @scroll      = opts.fetch(:scroll, '5m')
-        @size        = opts.fetch(:size, 50)
-        @search_type = opts.fetch(:search_type, nil)
+        default_opts = {
+          :index => nil,
+          :type => nil,
+          :scroll => '5m',
+          :size => 50,
+          :body => query,
+        }
```
grantr: I'd put these in a constant above the method like so:

```ruby
DEFAULT_OPTS = {
  :index  => nil,
  :type   => nil,
  :scroll => '5m',
  :size   => 50
}.freeze
```

I'd probably move the `:body` merge somewhere else, since it's already an ivar.

lib/elastomer/client/scroller.rb
```diff
@@ -128,19 +135,14 @@ class Scroller
...
-        @size        = opts.fetch(:size, 50)
-        @search_type = opts.fetch(:search_type, nil)
+        @opts = DEFAULT_OPTS.merge({ :body => query }).merge(opts)
 
         @scroll_id = nil
         @offset = 0
       end
 
-      attr_reader :client, :query, :index, :type, :scroll, :size, :search_type, :scroll_id
+      attr_reader :client, :query, :scroll_id
```
grantr: Did you check to see if there are any uses of these reader methods? We may want to continue supporting them to avoid breakage.

test/client/scroller_test.rb
```diff
@@ -83,6 +83,16 @@
...
+  it 'propagates URL query strings' do
+    scan = @index.scan(nil, { :q => 'author:pea53 || title:17' })
+
+    counts = {'tweet' => 0, 'book' => 0}
+    scan.each_document { |h| counts[h['_type']] += 1 }
+
+    assert_equal 50, counts['tweet']
+    assert_equal 1, counts['book']
+  end
+
```
grantr: Thanks for writing a test :smiley:

lib/elastomer/client/scroller.rb
chrismwendt: Good catch - I forgot that they're visible outside of the class. I checked and I don't see any references to them elsewhere in the library. If it's likely that this is a user-facing breaking change, they could be implemented by reading from the `@opts` hash, or by putting `#scroll_opts` back. I'll defer to your judgement here.

grantr: Looks like nothing outside this library uses Scroller directly, so I think we're ok to remove. Still, we should note this as a breaking change in CHANGELOG.md and increment version numbers appropriately.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0 (2015-09-11)
+- Support all URL parameters when using `Client.#scroll`
+- BREAKING: Moved some `Scroller` reader methods into `Scroller.opts`
```
grantr: This is very nitpicky, but I tend to put breaking changes at the top of the changes list.

Issue #104, opened 2015-08-18
Implement delete_by_query as scan and bulk delete


grantr:
The `delete_by_query` API is going away in ES 2.0 because of multiple issues (See https://www.elastic.co/blog/core-delete-by-query-is-a-plugin). The alternative is a combination scan and bulk delete.

We should make the change in elastomer-client now to prep for the ES 2.0 upgrade.

/cc @TwP

chrismwendt:
Should the command be renamed as well? Doing so would avoid some confusion like "I thought this API was removed in 2.0?", but it would be more work to find all references to `delete_by_query` and rename them.

grantr:
@chrismwendt Good question. Can we gather some data to find out how often this command is used? I suspect it's mostly used in test/dev code. Here's a search for it: https://github.com/search?p=1&q=%40github+delete_by_query&type=Code&utf8=%E2%9C%93

I can't think of a better name than `delete_by_query`, can you?

chrismwendt:
It's hard to say how much it's used, and I can't think of a better name either. Perhaps it's appropriate to leave the name the same because the implementation, not the functionality, will change.

TwP:
My vote is to leave the `delete_by_query` name in place. The scan / bulk delete approach will word on versions of ES prior to 2.0, and this approach is *safer* than the current `delete_by_query` implementation in ES itself. So hiding this endpoint from the user is a good, opinionated change to make. We should absolutely update the method documentation to let people know what is going on under the hood.

chrismwendt:
Implemented in #108.

PR #108, opened 2015-09-11
Reimplement delete_by_query using scroll and bulk delete

chrismwendt:
This changes the implementation of `delete_by_query` from the Delete by Query API, which was removed in ES 2.0, to scrolling combined with bulk deletion **and** changes the return value format to something similar to what's returned by the plugin version of Delete by Query: https://github.com/elastic/elasticsearch/blob/master/docs/plugins/delete-by-query.asciidoc#response-body.

The accompanying test will not pass until #106 is merged in.

See #104.


lib/elastomer/client/docs.rb
```diff
@@ -269,8 +269,40 @@ def count( query, params = nil )
...
+            "deleted" => items.count { |i| is_ok.call(i["status"]) },
+            "missing" => items.count { |i| !i["found"] },
+            "failed" => items.count { |i| i["found"] && !is_ok.call(i["status"]) },
+          }
+        end
+
+        indices = response_items
+          .map { |i| { i["_index"] => [i.tap { |ip| ip.delete("_index") }] } }
+          .reduce({}) { |acc, i| acc.merge(i) { |_, old, new| old + new } }
+          .map { |index, items| [index, categorize.call(items)] }.to_h
```
chrismwendt: This is moderately complicated data transformation, so let me know if and where I should add comments.

lib/elastomer/client/docs.rb
```diff
@@ -265,12 +272,44 @@ def count( query, params = nil )
...
         query, params = extract_params(query) if params.nil?
 
-        response = client.delete '/{index}{/type}/_query', update_params(params, :body => query, :action => 'docs.delete_by_query')
-        response.body
+        responses = []
+        accumulate = lambda { |response| responses << response unless response == nil }
+
+        accumulate.call(bulk do |b|
+          scan(query, params).each_document do |hit|
+            accumulate.call(b.delete(_id: hit["_id"]))
```
grantr: This fails for Docs objects without a given type because the `_type` isn't automatically filled in:

```
Elastomer::Client::RequestError: ActionRequestValidationException[Validation Failed: 1: type is missing;2: type is missing;3: type is missing;4: type is missing;]
```

To avoid that, always pass in the `_type`. It's probably also a good idea to pass in the `_index` as well in case it's possible to call delete_by_query without index scope.

```ruby
b.delete(_id: hit["_id"], _type: hit["_type"], _index: hit["_index"])
```

Another wrinkle: Indices with document routing enabled require the `_routing` parameter to delete. You'll probably want to check for that in the scan response and pass it in.

lib/elastomer/client/docs.rb
```diff
@@ -265,12 +272,44 @@ def count( query, params = nil )
...
+          end
+        end)
+
+        response_items = responses.flat_map { |r| r["items"].map { |i| i["delete"] } }
```
grantr: Collects the `delete` action hashes into an array

lib/elastomer/client/docs.rb
```diff
@@ -265,12 +272,44 @@ def count( query, params = nil )
...
+          {
+            "found" => items.count { |i| i["found"] },
+            "deleted" => items.count { |i| is_ok.call(i["status"]) },
+            "missing" => items.count { |i| !i["found"] },
+            "failed" => items.count { |i| i["found"] && !is_ok.call(i["status"]) },
+          }
+        end
+
+        indices = Hash[response_items
+          .map { |i| { i["_index"] => [i.tap { |ip| ip.delete("_index") }] } }
```
grantr: Removes the `_index` key from the delete hashes

lib/elastomer/client/docs.rb
```diff
@@ -265,12 +272,44 @@ def count( query, params = nil )
...
+          .reduce({}) { |acc, i| acc.merge(i) { |_, old, new| old + new } }
```
grantr: Groups the delete hashes by index name

lib/elastomer/client/docs.rb
chrismwendt: Are you sure that `_routing` is required for deletes? This says it's optional: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete.html#delete-routing

Even if we wanted to provide the `_routing` parameter (in order to avoid the delete operation being broadcast to all shards), I don't think we can in general because the routing parameter that was used to index the document is not always stored. It's only retrievable when routing is read from a field in the document, but that's deprecated https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-routing-field.html#_path_2.

lib/elastomer/client/docs.rb
grantr: [Enumerable#group_by](http://ruby-doc.org/core-1.9.3/Enumerable.html#method-i-group_by) might be useful. Could possibly replace this line and the previous.

grantr:
@chrismwendt I left a few line comments with my understanding of what the line is doing. Please correct me if I'm wrong!

The length of the `delete_by_query` method and the number of lambda functions required suggest that this might be better implemented as a command object, similar to Bulk. I think that would be easier to follow and more idiomatic Ruby (at least, idiomatic Ruby at GitHub).

lib/elastomer/client/docs.rb
grantr: Ah good point, I didn't see that it's optional. Since it's not retrievable we shouldn't send it here.

lib/elastomer/client/docs.rb
grantr: I'd add a comment above each line explaining what its purpose is. I'd also comment on why accumulate is called both inside and outside the bulk block, and what the purpose of the `flat_map` line is (since `flat_map` is rarely used)

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,127 @@
...
+      #
+      # item - An element of the items array from a bulk response
+      #
+      # Returns a Hash of counts for each category
+      def categorize(item)
+        {
+          "found" => item["found"] || item["status"] == 409 ? 1 : 0,
+          "deleted" => is_ok?(item["status"]) ? 1 : 0,
+          "missing" => !item["found"]  && !item.key?("error") ? 1 : 0,
+          "failed" => item.key?("error") ? 1 : 0,
```
chrismwendt: Even with access to the [docs](https://github.com/elastic/elasticsearch/blob/master/docs/plugins/delete-by-query.asciidoc), [source](https://github.com/elastic/elasticsearch/tree/master/plugins/delete-by-query), and Elasticsearch itself, it's difficult to make sure that these checks are right in all cases. I ran some tests with Elasticsearch with some infinite bash loops to cause version conflicts and missing documents, and these checks seem to be consistent with the responses I got from the official Delete by Query plugin.

Let me know how watertight this should be.

grantr: This is really close. I'd like to see a test that checks whether counts are aggregated correctly. It's fine to demonstrate this only with deletes; no need for missing or failure aggregation checks.

test/client/delete_by_query_test.rb
```diff
@@ -0,0 +1,136 @@
...
+      @index.refresh
+
+      count = 0
+      WebMock.after_request do |request, _|
+        count += 1 if request.uri.path =~ /_bulk$/
+      end
+
+      response = $client.delete_by_query(nil, :action_count => 1)
+
+      assert_equal(2, count)
```
grantr: Add an assertion here that response has `found` and `deleted` equal to 2.

grantr: Also `refute_found` for both docs.

grantr:
I'm :+1: to merge this once the test additions are in. Good work @chrismwendt!

grantr:
Looks great @chrismwendt! I say :ship:

chrismwendt:
Thanks!

PR #109, opened 2015-09-14
Translate the delete_by_query method into a command object

chrismwendt:
This breaks up the rather large delete_by_query method and puts it in its own class so that it's more manageable. Note that is a branch off of use-scroll-for-delete-by-query, not master.

Command objects are new to me, so this probably isn't the proper way to do it. Let me know if you have any feedback.

/cc @TwP @grantr

### DIFF
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -57,7 +57,8 @@ module Elastomer
         unless response.nil?
           @response_stats['took'] += response['took']
 
-          response['items'].map { |i| i['delete'] }.each do |i|
+          response['items'].each do |i|
+            i = i['delete']
             (@response_stats['_indices'][i['_index']] ||= {}).merge!(categorize(i)) { |_, n, m| n + m }
             @response_stats['_indices']['_all'].merge!(categorize(i)) { |_, n, m| n + m }
             @response_stats['failures'] << i unless is_ok? i['status']
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -54,7 +54,7 @@ module Elastomer
       end
 
       def accumulate(response)
-        unless response == nil
+        unless response.nil?
           @response_stats['took'] += response['took']
 
           response['items'].map { |i| i['delete'] }.each do |i|
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -29,27 +29,33 @@ module Elastomer
         @responses << response unless response == nil
       end
 
-      # Perform bulk indexing and/or delete operations. The current index name
-      # and document type will be passed to the bulk API call as part of the
-      # request parameters.
+      # Delete documents from one or more indices and one or more types based
+      # on a query. This method supports both the "request body" query and the
+      # "URI request" query. When using the request body semantics, the query
+      # hash must contain the :query key. Otherwise we assume a URI request is
+      # being made.
       #
-      # params - Parameters Hash that will be passed to the bulk API call.
-      # block  - Required block that is used to accumulate bulk API operations.
-      #          All the operations will be passed to the search cluster via a
-      #          single API request.
+      # The return value follows the format returned by the Elasticsearch Delete
+      # by Query plugin: https://github.com/elastic/elasticsearch/blob/master/docs/plugins/delete-by-query.asciidoc#response-body
       #
-      # Yields a Bulk instance for building bulk API call bodies.
+      # Internally, this method uses a combination of scroll and bulk delete
+      # instead of the Delete by Query API, which was removed in Elasticsearch
+      # 2.0.
+      #
+      # query  - The query body as a Hash
+      # params - Parameters Hash
       #
       # Examples
       #
-      #   docs.bulk do |b|
-      #     b.index( document1 )
-      #     b.index( document2 )
-      #     b.delete( document3 )
-      #     ...
-      #   end
+      #   # request body query
+      #   delete_by_query({:query => {:match_all => {}}}, :type => 'tweet')
       #
-      # Returns the response body as a Hash
+      #   # same thing but using the URI request method
+      #   delete_by_query(:q => '*:*', :type => 'tweet')
+      #
+      # See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
+      #
+      # Returns a Hash of statistics about the delete operations
       def execute()
         # accumulate is called both inside and outside the bulk block in order
         # to capture bulk responses returned from calls to `delete` and the call
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -57,11 +57,11 @@ module Elastomer
         unless response.nil?
           @response_stats['took'] += response['took']
 
-          response['items'].each do |i|
-            i = i['delete']
-            (@response_stats['_indices'][i['_index']] ||= {}).merge!(categorize(i)) { |_, n, m| n + m }
-            @response_stats['_indices']['_all'].merge!(categorize(i)) { |_, n, m| n + m }
-            @response_stats['failures'] << i unless is_ok? i['status']
+          response['items'].each do |item|
+            item = item['delete']
+            (@response_stats['_indices'][item['_index']] ||= {}).merge!(categorize(item)) { |_, n, m| n + m }
+            @response_stats['_indices']['_all'].merge!(categorize(item)) { |_, n, m| n + m }
+            @response_stats['failures'] << item unless is_ok? item['status']
           end
         end
       end
@@ -70,9 +70,9 @@ module Elastomer
         # accumulate is called both inside and outside the bulk block in order
         # to capture bulk responses returned from calls to `delete` and the call
         # to `bulk`
-        accumulate(@client.bulk(@params) do |b|
+        accumulate(@client.bulk(@params) do |bulk|
           @client.scan(@query, @params).each_document do |hit|
-            accumulate(b.delete(_id: hit["_id"], _type: hit["_type"], _index: hit["_index"]))
+            accumulate(bulk.delete(_id: hit["_id"], _type: hit["_type"], _index: hit["_index"]))
           end
         end)
 
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -109,7 +109,7 @@ module Elastomer
       # Perform the Delete by Query action
       #
       # Returns a Hash of statistics about the bulk operation
-      def execute()
+      def execute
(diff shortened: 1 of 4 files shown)


lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,82 @@
...
+      def initialize( client, query, params = nil )
+        @client    = client
+        @query     = query
+        @params    = params
+        @responses = []
+      end
+
+      attr_reader :client, :query, :params, :responses
+
+      def is_ok( status )
```
grantr: Normally methods that return a boolean value are suffixed with a `?`, e.g. `is_ok?`

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,82 @@
+module Elastomer
+  class Client
+
+    class DeleteByQuery
+
+      def initialize( client, query, params = nil )
```
grantr: The style guide specifies no spaces between parentheses in method definitions, the rest of this library notwithstanding (cc @TwP the scofflaw)

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,82 @@
...
+      # Examples
+      #
+      #   docs.bulk do |b|
+      #     b.index( document1 )
+      #     b.index( document2 )
+      #     b.delete( document3 )
+      #     ...
+      #   end
+      #
+      # Returns the response body as a Hash
```
grantr: Docs here are for `Bulk`, might want to correct those

lib/elastomer/client/docs.rb
```diff
@@ -275,45 +275,9 @@ def count( query, params = nil )
       # Returns a Hash of statistics about the delete operations
       def delete_by_query( query, params = nil )
         query, params = extract_params(query) if params.nil?
+        params = { :index => @name, :type => @type }.merge params
```
grantr: Use the `update_params` method here, as in https://github.com/github/elastomer-client/blob/command-object/lib/elastomer/client/docs.rb#L294

lib/elastomer/client/docs.rb
```diff
@@ -275,45 +275,9 @@ def count( query, params = nil )
...
-          .map { |index, items| [index, categorize.call(items)] }]
-
-        indices_with_all = indices.merge({ "_all" => indices.values.reduce({}) { |acc, i| acc.merge(i) { |_, n, m| n + m } } })
-
-        {
-          "took" => responses.map { |r| r["took"] }.reduce(:+),
-          "_indices" => indices_with_all,
-          "failures" => response_items.select { |i| !is_ok.call(i["status"]) },
-        }
+        DeleteByQuery.new(@client, query, params).execute()
```
grantr: No need for the empty parens here

grantr:
@chrismwendt The Bulk helper defines the `bulk` method in its own file: https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/bulk.rb#L31

Can you explain why you aren't using this style for `DeleteByQuery`? (That's not a trick question btw, I'm curious about your reasoning)

chrismwendt:
There really isn't a reason why I didn't use that style for `DeleteByQuery`, thanks for pointing that out. I'll make it consistent with `Bulk`.

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,82 @@
...
+        {
+          "found" => items.count { |i| i["found"] },
+          "deleted" => items.count { |i| is_ok(i["status"]) },
+          "missing" => items.count { |i| !i["found"] },
+          "failed" => items.count { |i| i["found"] && !is_ok(i["status"]) },
+        }
+      end
+
+      def accumulate( response )
+        @responses << response unless response == nil
```
TwP: In chat we already talked about the response accumulation here. I'm looking forward to the improvements to the bulk operations that will make this process unneeded.

Here is an arcane bit of ruby trivia: `response.nil?` is faster to process than `response == nil`. I prefer the interrogatory methods over equality checks for this reason. And it is more difficult to do the dreaded single `=` mistake `response = nil` in an if statement.

lib/elastomer/client/delete_by_query.rb
chrismwendt: Ok, I will remove those spaces. Shall we make a sweep through the whole repo with a whitespace normalization commit as a separate PR?

TwP: Definitely a separate PR for that sweep. And the guilty party should be the one writing that PR as penance.

lib/elastomer/client/delete_by_query.rb
chrismwendt: I will commit the incremental statistics in just a minute, stay tuned.

Good point on the `=` vs `==` mistake - I'll be sure to use the interrogatory form.

lib/elastomer/client/docs.rb
```diff
@@ -273,47 +273,10 @@ def count( query, params = nil )
...
-          .map { |index, items| [index, categorize.call(items)] }]
-
-        indices_with_all = indices.merge({ "_all" => indices.values.reduce({}) { |acc, i| acc.merge(i) { |_, n, m| n + m } } })
-
-        {
-          "took" => responses.map { |r| r["took"] }.reduce(:+),
-          "_indices" => indices_with_all,
-          "failures" => response_items.select { |i| !is_ok.call(i["status"]) },
-        }
+        @client.delete_by_query(query, update_params(params, { :index => @name, :type => @type }))
```
grantr: No need to override `:index` and `:type` here, they'll be added by `Docs#defaults`.

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,83 @@
...
+          "missing" => !item["found"] ? 1 : 0,
+          "failed" => item["found"] && !is_ok?(item["status"]) ? 1 : 0,
+        }
+      end
+
+      def accumulate(response)
+        unless response.nil?
+          @response_stats['took'] += response['took']
+
+          response['items'].map { |i| i['delete'] }.each do |i|
```
TwP: I would skip the `map` and pull out the `i["delete"]` inside the `each`. Iterating twice over the results will not be good for performance.

```ruby
response["items"].each do |item|
  item = item["delete"]
  ...
end
```

One habit I *have* been broken of is using single quotes. We prefer to use double quotes when working with strings.

Single character variable names are verboten :no_entry_sign:

grantr: > Single character variable names are verboten

https://github.com/github/elastomer-client/blame/master/lib/elastomer/client/docs.rb#L536-L539 :stuck_out_tongue:

TwP: */me hangs head in shame*

Adds this to the list of refactorings as part of my penance.

chrismwendt: There are so many `'`s that I'll save them for a repo-wide sweep later.

grantr: I'm neutral on the `'` => `"` swap. Sometimes I prefer to preserve blame rather than fix a style issue. (Related: I wish we had a good blame explorer mode)

chrismwendt: Sounds like something that a semantic diff would be able to handle :smile: /cc @robrix

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,84 @@
...
+    #
+    #   # request body query
+    #   delete_by_query({:query => {:match_all => {}}}, :type => 'tweet')
+    #
+    #   # same thing but using the URI request method
+    #   delete_by_query(nil, { :q => '*:*', :type => 'tweet' })
+    #
+    # See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
+    #
+    # Returns a Hash of statistics about the delete operations
```
grantr: Can we document an example of what this hash looks like?

lib/elastomer/client/docs.rb
```diff
@@ -273,47 +273,10 @@ def count( query, params = nil )
       # See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
       #
       # Returns a Hash of statistics about the delete operations
```
grantr: Ok to refer to the documentation for `Client#delete_by_query` here.

lib/elastomer/client/index.rb
```diff
@@ -523,6 +523,34 @@ def warmer( warmer_name )
...
+      #
+      #   # request body query
+      #   delete_by_query({:query => {:match_all => {}}}, :type => 'tweet')
+      #
+      #   # same thing but using the URI request method
+      #   delete_by_query(:q => '*:*', :type => 'tweet')
+      #
+      # See http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
+      #
+      # Returns a Hash of statistics about the delete operations
```
grantr: Ok to refer to the documentation for `Client#delete_by_query` here.

test/client/index_test.rb
```diff
@@ -362,5 +362,25 @@
...
+          'failed' => 0,
+        },
+        @name => {
+          'found' => 1,
+          'deleted' => 1,
+          'missing' => 0,
+          'failed' => 0,
+        }
+      }, r['_indices'])
+    end
```
grantr: I'd like to see some tests for the missing/failed scenarios, but since webmock is added in #108 it can wait until after merge.

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+      def categorize(item)
+        {
+          "found" => item["found"] ? 1 : 0,
+          "deleted" => is_ok?(item["status"]) ? 1 : 0,
+          "missing" => !item["found"] ? 1 : 0,
+          "failed" => item["found"] && !is_ok?(item["status"]) ? 1 : 0,
+        }
+      end
```
grantr: These instance methods should be documented. They are all `# Internal:` except `initialize` and `execute`. See http://tomdoc.org/

lib/elastomer/client/delete_by_query.rb
```diff
@@ -0,0 +1,84 @@
...
+          response['items'].each do |item|
+            item = item['delete']
+            (@response_stats['_indices'][item['_index']] ||= {}).merge!(categorize(item)) { |_, n, m| n + m }
+            @response_stats['_indices']['_all'].merge!(categorize(item)) { |_, n, m| n + m }
+            @response_stats['failures'] << item unless is_ok? item['status']
+          end
+        end
+      end
+
+      def execute()
```
grantr: No empty parentheses here. Just `def execute` for methods that take no parameters.

grantr:
This is great @chrismwendt! I'm :+1: to merge.

TwP:
:sparkles: :shipit:

Issue #110, opened 2015-09-14
Style guide cleanup


TwP:
We are following GitHub's [ruby style guide](https://github.com/styleguide/ruby), but there are many places in the codebase where the style guide has been ignored:

* whitespace between parentheses and values
* single character variable names should be eschewed
* quotes `"good"` should be used for strings instead of apostrophes `'bad'`
* hashrockets `=>` should be used instead of the JSON style `:` in hashes
* trailing whitespace

Please add others to this list so I can do a sweep and fix my wayward coding style.

/cc @grantr @chrismwendt

chrismwendt:
I added the quotes, hashrockets, and trailing whitespace points since I have witnessed or committed such atrocities myself.

Perhaps we could establish a pre-commit hook to run a style check.

TwP:
[Rubocop](https://github.com/bbatsov/rubocop) seems to be the Ruby style checker de jour. We could throw that into a pre-push hook on the project much like puppet does for validating puppet files. Rubcop would run locally then before pushing changes up to the origin repository.

Good idea, @chrismwendt :clap: :+1: :sparkles:

chrismwendt:
#111 cleaned up all but the single-character variable names. @TwP: it's your call if you would like to close this issue.

TwP:
Single variable names are fine. :+1: on closing this out, and :bow: @chrismwendt for getting rubocop up an running on this repo.

PR #111, opened 2015-09-16
Set up RuboCop and Overcommit

chrismwendt:
This commit sets up style checking with [RuboCop](https://github.com/bbatsov/rubocop) and uses [Overcommit](https://github.com/brigade/overcommit) to manage hooks.

These are automatically set up by running `script/bootstrap`, so a new contributor doesn't need to run any extra commands to get started.

The purpose of `.rubocop_todo.yml` is to ignore for now offenses that will be resolved later (see https://github.com/bbatsov/rubocop#automatically-generated-configuration for the workflow).

Overcommit installs some hooks by default (e.g. max commit subject length) which we might want to disable, but I left them as-is because the ones I ran into are reasonable and simply result in warnings.

@TwP and @grantr: Please pull this branch down locally, run `script/bootstrap`, modify a random Ruby file, and make sure you can still commit. You can also try removing some of the entries in `.rubocop_todo.yml` to see what offenses are reported.

Whether or not RuboCop should be run as a branch check (like a Travis build) is an open question.

See the associated "Style guide cleanup" issue: #110.

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -2,4 +2,4 @@ PreCommit:
   RuboCop:
     enabled: true
     problem_on_unmodified_line: ignore
-    flags: ['--fail-level', 'error']
+    flags: ['--fail-level', 'error', '--format', 'emacs']
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -1,0 +1,5 @@
+PreCommit:
+  RuboCop:
+    enabled: true
+    problem_on_unmodified_line: ignore
+    flags: ['--fail-level', 'error']
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,81 @@
+# Ruby linting configuration.
+# We only worry about two kinds of issues: 'error' and anything less than that.
+# Error is not about severity, but about taste. Simple style choices that
+# never have a great excuse to be broken, such as 1.9 JSON-like hash syntax,
+# are errors. Choices that tend to have good exceptions in practice, such as
+# line length, are warnings.
+
+# If you'd like to make changes, a full list of available issues is at
+#   https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml
+# A list of configurable issues is at:
+#   https://github.com/bbatsov/rubocop/blob/master/config/default.yml
+#
+# If you disable a check, document why.
+
+
+StringLiterals:
+  EnforcedStyle: double_quotes
+  Severity: error
+
+HashSyntax:
+  EnforcedStyle: hash_rockets
+  Severity: error
+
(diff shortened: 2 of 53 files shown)


.overcommit.yml
```diff
@@ -0,0 +1,4 @@
+PreCommit:
```
chrismwendt: @TwP: I set this as a pre-commit hook in order to tighten the feedback loop. It can certainly be changed to a pre-push hook if you think that would be a better option.

chrismwendt:
An unfortunate side effect of hooks are that most GUIs fail to find the executables required. I'm open to suggestions on remedying that situation.

.rubocop_todo.yml
```diff
@@ -0,0 +1,33 @@
...
+    - 'lib/elastomer/middleware/opaque_id.rb'
+    - 'lib/elastomer/notifications.rb'
+    - 'test/client/docs_test.rb'
+
+# Offense count: 1372
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles.
+Style/StringLiterals:
+  Exclude:
+    - '**/*'
```
chrismwendt: There are so many single quote offenses that rubocop's config generation simply turned it off with `Enable: false` rather than ignoring each individual file. However, that same property was overridden in `.rubocop.yml` to get turned back on again `Enable: true`, so I instead set `Exclude: -'**/*'` to be consistent with the other cops.

TwP:
I'll throw some :eyes: on this tomorrow. A little bogged down with other tasks at the moment.

chrismwendt:
I just noticed github/github has a RuboCop config file - should we use that instead for consistency? https://github.com/github/github/blob/master/.rubocop.yml

TwP:
@chrismwendt :+1: for consistency with the main GitHub repository.

chrismwendt:
It's best to review this commit-by-commit, ignoring `Fix RuboCop Style/StringLiterals offenses` ad4f9aa, since that is a huge change which was done using `rubocop --auto-correct`.

.overcommit.yml
grantr: I prefer pre-commit because it avoids "fixing rubocop" commits in the history. We'll see if it gets too annoying though.

grantr:
One issue I've run into with pre-commit hooks is running them on merges when no code has been changed. Is overcommit smart enough to skip pre-commit on merges?

grantr:
This all looks good! :+1:

chrismwendt:
@grantr: Right now, Overcommit is configured to run RuboCop only as a pre-commit hook, so it won't be run when a merge happens.

Issue #92, opened 2014-12-05
Using action_count or request_size swallows bulk response


grantr:
Normally a bulk request returns the bulk response:

```
client.bulk do |bulk_request|
  bulk_request.index({"foo" => "bar"}, :id => '1', :index => 'test', :type => 'test')
end

=> {"took"=>61, "errors"=>false, "items"=>[{"index"=>{"_index"=>"test", "_type"=>"test", "_id"=>"1", "_version"=>1, "status"=>201}}]}
```

When using the `:action_count` or `:request_size` options to chunk bulk requests behind the scenes, the response is lost:

```
client.bulk(:action_count => 1) do |bulk_request|
  bulk_request.index({"foo" => "bar"}, :id => '1', :index => 'test', :type => 'test')
  bulk_request.index({"foo" => "bar"}, :id => '2', :index => 'test', :type => 'test')
end

=> nil
```

I think when `:action_count` or `:request_size` are used, the return value should be an array of response hashes.

I can't decide what the return value should be if no requests are made. The non-chunking `bulk` returns nil, but maybe the chunking version should return an empty array.

Is it worth making the multiple-request bulk helper a separate class? We use it heavily, so changing it would be painful, but it has opinions about how to handle multiple requests that maybe don't belong in the single-request bulk helper. Perhaps we could move it to a utility gem and supplement it with other tools like a standard scan/bulk looper.

chrismwendt:
Check out `bulk_stream_items` which was implemented in #112. https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/bulk.rb#L119

grantr:
Fixed by #112.

PR #112, opened 2015-09-17
Streaming bulk operations

chrismwendt:
This adds streaming bulk functionality via Ruby's Enumerators.

This does not to alter any of the existing methods in backwards incompatible ways so that a conversation about making the API more consistent could take place later. Because of that, some of the small operation methods had to be nearly duplicated.

This solves #92.

/cc @TwP @grantr

### DIFF
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -81,7 +81,8 @@ module Elastomer
       item.values.first["status"].between?(200, 299)
     end
 
-    # Stream bulk actions from an Enumerator and passes them to the given block.
+    # Stream bulk actions from an Enumerator and passes the response items to
+    # the given block.
     #
     # Examples
     #
@@ -100,6 +101,18 @@ module Elastomer
     #   #   "errors" => false,
     #   #   "success" => 3,
     #   #   "failure" => 0
+    #   # }
+    #
+    #   # sample response item:
+    #   # {
+    #   #   "delete": {
+    #   #     "_index": "foo",
+    #   #     "_type": "bar",
+    #   #     "_id": "42",
+    #   #     "_version": 3,
+    #   #     "status": 200,
+    #   #     "found": true
+    #   #   }
     #   # }
     #
     # Returns a Hash of stats about items from the responses.
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -381,8 +381,8 @@ module Elastomer
       # size, then a bulk API call will be performed. After the call the
       # actions list is cleared and we'll start accumulating actions again.
       #
-      # action - The bulk action (as a Hash) to perform
-      # document  - Optional document for the action as a Hash or JSON encoded String
+      # action   - The bulk action (as a Hash) to perform
+      # document - Optional document for the action as a Hash or JSON encoded String
       #
       # Returns the response from the bulk call if one was made or nil.
       def add_to_actions( action, document = nil )
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -71,19 +71,61 @@ module Elastomer
       end
     end
 
-    # Stream bulk actions from an Enumerator.
+    # Internal: Determine whether or not a response item has an HTTP status code
+    # in the range 200 to 299.
     #
-    # See Client#bulk_stream_responses for more information.
+    # item - The bulk response item
     #
-    # Returns an Enumerator of items from the responses.
+    # Returns a boolean
+    def is_ok?(item)
+      item.values.first["status"].between?(200, 299)
+    end
+
+    # Stream bulk actions from an Enumerator and passes them to the given block.
+    #
+    # Examples
+    #
+    #   ops = [
+    #     Bulk::index(document1, :_type => 'default-type', :_id => 1),
+    #     Bulk::index(document2, params2),
+    #     Bulk::delete(params3),
+    #   ]
+    #   bulk_stream_items(ops, :index => 'default-index') do |item|
+    #     puts item
+    #   end
+    #
+    #   # return value:
+    #   # {
+    #   #   "took" => 256,
+    #   #   "errors" => false,
+    #   #   "success" => 3,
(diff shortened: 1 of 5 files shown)


lib/elastomer/client/bulk.rb
```diff
@@ -43,6 +43,48 @@ def bulk( body = nil, params = nil )
...
+    # Returns an Enumerator of items from the responses.
+    def bulk_stream_items(ops, params = {})
+      Enumerator.new do |yielder|
+        bulk_stream_responses(ops, params).each do |response|
+          response["items"].each do |item|
+            yielder.yield item
+          end
+        end
+      end
+    end
```
chrismwendt: Is there a simpler way to essentially `map` over an `Enumerator` than by constructing a new one and yielding each element of the original, as I did here?

TwP: On thought on this method is to **not** return an `Enumerator`. Instead, accept a block and perform the iteration directly with this method.

```ruby
def bulk_stream_items(ops, params = {})
  stats = {
    "took" => 0,
    "errors" => false,
    "success" => 0,
    "failure" => 0
  }

  bulk_stream_responses(ops, params).each do |response|
    stats["took"] += response["took"]
    stats["errors"] |= response["errors"]

    response["items"].each do |item|
      if is_ok?(item)
        stats["success"] += 1
      else
        stats["failure"] += 1
      end
      yield item
    end
  end

  return stats
end
```

I've also added a hash of interesting numbers as the return value from this method. The `is_ok?` method does not exist yet, but it would check the HTTP status returned with each item.

```
def is_ok?(item)
  item["status"].between?(200, 299)
end
```

Thoughts on this approach?

chrismwendt: The client might want to retry failed documents, in which case they would need access to at least the failed response items. Maybe it's worth having a separate convenient method called `bulk_stream_stats` which does exactly what you suggest.

TwP: The client does have access to the failed items during the iteration. When `yield item` is called, the client can also check the status code of that particular operation.

Returning the stats hash is there just to be nice to the end user.

lib/elastomer/client/bulk.rb
```diff
@@ -43,6 +43,48 @@ def bulk( body = nil, params = nil )
...
 
+    # Stream bulk actions from an Enumerator.
+    #
+    # Examples
+    #
+    #   ops = [
+    #     Bulk::index(document1, :_type => 'default-type', :_id => 1),
+    #     Bulk::index(document2, params2),
+    #     Bulk::delete(params3),
+    #   ]
```
TwP: I'm not a huge fan of the `Bulk::index` / `Bulk::delete` syntax here. My proposal is this:

```ruby
ops = [
  [:index, document1, {:_type => "foo", :_id => 1}],
  [:create, document2],
  [:delete, {:_type => "bar", :_id => 42}]
]
```

The first item in the list is the bulk action to take. This corresponds to one of the action methods on the Bulk instance. The second item in the list is a document (or the params in the case of `:delete`). And thirdly comes the optional params.

Inside the Enumerator we have:

```ruby
Enumerator.new do |yielder|
  ops.each do |action, *args|
    response = bulk_obj.send(action, *args)
    yielder.yield response unless response.nil?
  end
  response = bulk_obj.call
  yielder.yield response unless response.nil?
end
```

The Ruby `send` method is being used here to call the `action` method on the `bulk_obj` and passing the `*args` to the method.

lib/elastomer/client/bulk.rb
chrismwendt: My apologies, I read over your code too quickly and missed that `yield item` :bow:

I like the convenience of returning stats, and the block syntax would look almost the same as calling `each` on the `Enumerator`. I'll implement it like you showed.

lib/elastomer/client/bulk.rb
chrismwendt: That's a pretty clever use of `send` to reuse the existing methods! I'll implement that now.

lib/elastomer/client/bulk.rb
```diff
@@ -211,7 +295,7 @@ def call
 
       # Internal: convert special key parameters to their wire representation
       # and apply any override document parameters.
-      def prepare_params(document, params)
+      def self.prepare_params(document, params)
```
TwP: This can be reverted back to an instance method.

lib/elastomer/client/bulk.rb
```diff
@@ -227,7 +311,7 @@ def prepare_params(document, params)
       # document - The document Hash
       #
       # Returns extracted key/value pairs as a Hash.
-      def from_document( document )
+      def self.from_document( document )
```
TwP: This can be reverted back to an instance method.

lib/elastomer/client/bulk.rb
```diff
@@ -249,7 +333,7 @@ def from_document( document )
       # params - Hash.
       #
       # Returns a new params Hash with the special keys replaced.
-      def convert_special_keys(params)
+      def self.convert_special_keys(params)
```
TwP: This can be reverted back to an instance method.

lib/elastomer/client/bulk.rb
```diff
@@ -43,6 +43,90 @@ def bulk( body = nil, params = nil )
...
+    #
+    # Examples
+    #
+    #   ops = [
+    #     [:index, document1, {:_type => "foo", :_id => 1}],
+    #     [:create, document2],
+    #     [:delete, {:_type => "bar", :_id => 42}]
+    #   ]
+    #   bulk_stream_items(ops, :index => 'default-index') do |item|
+    #     puts item
```
grantr: Can we document what the item typically looks like?

chrismwendt:
@TwP: I reverted those class methods back to instance methods.

@grantr: I added a sample response item to the docs.

I also bumped the version and updated the changelog.

TwP:
:+1: :shipit: :sparkles: and that's enough emoji for now.

grantr:
This is great stuff, useful and easy to use. :+1:

Issue #24, opened 2013-08-20
Percolate API


grantr:
Percolate is not supported yet, but since it's changing completely in 1.0, maybe we should wait until that is released.

A friend of mine using percolate in production describes it as an ops nightmare currently, so I doubt we'll be needing it until the 1.0 revamp.

grantr:
:tada:

Issue #32, opened 2013-10-29
Attempting to delete a document with id nil deletes the type or index


grantr:
We should tighten our url generation so that nil parameters can't accidentally delete higher level objects.

```
>> es = Elastomer::Client.new(:url => "http://localhost:19200")
=> #<Elastomer::Client:0x007f91e31e1500 @url="http://localhost:19200", @host="localhost", @port=19200, @read_timeout=4, @open_timeout=2, @adapter=:excon>
>> es.index('hello').create({})
=> {"ok"=>true, "acknowledged"=>true}
>> es.index('hello').exists?
=> true
>> es.index('hello').docs.delete(:id => nil)
=> {"ok"=>true, "acknowledged"=>true}
>> es.index('hello').exists?
=> false
```

PR #113, opened 2015-09-18
Implement the percolate API

chrismwendt:
This looked like low hanging fruit, so I decided to give it a shot. "Percolate API" #24 is the motivating issue.

- [x] Create/Retrieve/Delete/Exists on percolators
- [x] Percolate new document
- [x] Percolate existing document
- [x] Percolate count new and existing [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_percolator_count_api)
- [x] Multi percolate [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_multi_percolate_api)

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -267,7 +267,7 @@ module Elastomer
       #
       # Examples
       #
-      #   index.register_percolator_query 1, :query => { :match => { :author => "pea53" } }
+      #   index.percolator(1).create :query => { :match => { :author => "pea53" } }
       #   docs.percolate :doc => { :author => "pea53" }
       #   docs.percolate nil, :id => 3
       #
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -261,17 +261,19 @@ module Elastomer
         @client.delete_by_query(query, update_params(params))
       end
 
-      # Matches a document to the queries stored in the percolator for the given
-      # index.
+      # Matches a provided or existing document to the stored percolator
+      # queries. To match an existing document, pass `nil` as the body and
+      # include `:id` in the params.
       #
       # Examples
       #
       #   index.register_percolator_query 1, :query => { :match => { :author => "pea53" } }
       #   docs.percolate :doc => { :author => "pea53" }
+      #   docs.percolate nil, :id => 3
       #
       # Returns the response body as a Hash
-      def percolate(body, params = {})
-        response = client.get '/{index}/{type}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
+      def percolate(body = nil, params = {})
+        response = client.get '/{index}/{type}{/id}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
         response.body
       end
 
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -519,6 +519,31 @@ Percolate
         client.multi_search params, &block
       end
 
+      # Execute an array of percolate actions in bulk. Results are returned in
+      # an array in the order the actions were sent. The current index name and
+      # type will be passed to the API call as part of the request parameters.
+      #
+      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_multi_percolate_api
+      #
+      # params - Optional request parameters as a Hash
+      # block  - Passed to a MultiPercolate instance which assembles the
+      #          percolate actions into a single request.
+      #
+      # Examples
+      #
+      #   # block form
+      #   multi_percolate() do |m|
+      #     m.percolate({}, { :author => "pea53" })
+      #     m.count({}, { :author => "grantr" })
+      #     ...
+      #   end
+      #
+      # Returns the response body as a Hash
+      def multi_percolate(params = {}, &block)
+        params = { :index => self.name, :type => self.type }.merge params
+        client.multi_percolate(params, &block)
+      end
+
       SPECIAL_KEYS = %w[index type id version version_type op_type routing parent timestamp ttl consistency replication refresh].freeze
       SPECIAL_KEYS_HASH = SPECIAL_KEYS.inject({}) { |h, k| h[k.to_sym] = "_#{k}"; h }.freeze
 
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -272,7 +272,7 @@ module Elastomer
       #   docs.percolate nil, :id => 3
       #
       # Returns the response body as a Hash
-      def percolate(body = nil, params = {})
+      def percolate(body, params = {})
         response = client.get '/{index}/{type}{/id}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
         response.body
       end
@@ -288,7 +288,7 @@ module Elastomer
       #   docs.percolate_count nil, :id => 3
       #
       # Returns the count
-      def percolate_count(body = nil, params = {})
+      def percolate_count(body, params = {})
         response = client.get '/{index}/{type}{/id}/_percolate/count', update_params(params, :body => body, :action => 'percolator.percolate_count')
         response.body["total"]
       end
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -532,7 +532,7 @@ Percolate
       # Examples
       #
       #   # block form
-      #   multi_percolate() do |m|
+      #   multi_percolate do |m|
       #     m.percolate(:author => "pea53")
       #     m.count(:author => "grantr")
       #     ...
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -261,6 +261,21 @@ module Elastomer
         @client.delete_by_query(query, update_params(params))
       end
 
+      # Matches a document to the queries stored in the percolator for the given
+      # index. Type is required.
+      #
+      # Examples
+      #
+      #   index.register_percolator_query 1, { :query => { :match => { :author => "pea53" } } }
+      #   index.docs(type).percolate({ :doc => { :author => "pea53" } })
+      #
+      # Returns the response body as a Hash
+      def percolate(body, params = {})
+        raise 'percolate requires document type' if type.nil?
+        response = client.get '/{index}/{type}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
+        response.body
+      end
+
       # Returns information and statistics on terms in the fields of a
       # particular document as stored in the index. The :id is provided as part
       # of the params hash.
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -533,8 +533,8 @@ Percolate
       #
       #   # block form
       #   multi_percolate() do |m|
-      #     m.percolate({}, { :author => "pea53" })
-      #     m.count({}, { :author => "grantr" })
+      #     m.percolate(:author => "pea53")
+      #     m.count(:author => "grantr")
       #     ...
       #   end
       #
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -540,7 +540,7 @@ Percolate
       #
       # Returns the response body as a Hash
       def multi_percolate(params = {}, &block)
-        params = { :index => self.name, :type => self.type }.merge params
+        params = defaults.merge params
         client.multi_percolate(params, &block)
       end
 
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -262,12 +262,12 @@ module Elastomer
       end
 
       # Matches a document to the queries stored in the percolator for the given
-      # index. Type is required.
+      # index.
       #
       # Examples
       #
-      #   index.register_percolator_query 1, { :query => { :match => { :author => "pea53" } } }
-      #   index.docs(type).percolate({ :doc => { :author => "pea53" } })
+      #   index.register_percolator_query 1, :query => { :match => { :author => "pea53" } }
+      #   docs.percolate :doc => { :author => "pea53" }
       #
       # Returns the response body as a Hash
       def percolate(body, params = {})
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -271,7 +271,6 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def percolate(body, params = {})
-        raise 'percolate requires document type' if type.nil?
         response = client.get '/{index}/{type}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
         response.body
       end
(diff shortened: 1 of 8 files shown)


lib/elastomer/client/docs.rb
```diff
@@ -261,6 +261,21 @@ def delete_by_query(query, params = nil)
...
+      # index. Type is required.
+      #
+      # Examples
+      #
+      #   index.register_percolator_query 1, { :query => { :match => { :author => "pea53" } } }
+      #   index.docs(type).percolate({ :doc => { :author => "pea53" } })
+      #
+      # Returns the response body as a Hash
+      def percolate(body, params = {})
+        raise 'percolate requires document type' if type.nil?
```
grantr: `:type` can be passed in as a parameter so we don't know for sure if it's missing yet. I believe we normally allow the URI template to catch missing path segments.

test/client/docs_test.rb
```diff
@@ -518,6 +518,19 @@
       assert docs
       assert_equal(%w[1 2], docs.map { |h| h['_id'] }.sort)
     end
+
+    it 'percolates' do
+      populate!
+
+      response = @index.register_percolator_query "1", { :query => { :match => { :author => "pea53" } } }
+      assert response["created"]
```
grantr: For generic asserts, I like to add a message to describe what went wrong (the default message is "Assertion failed" or something equally useless).

```ruby
assert response["created"], "Couldn't create the percolator query"
```

lib/elastomer/client/docs.rb
chrismwendt: Ah, good catch. I'll remove the `raise` then.

test/client/docs_test.rb
chrismwendt: Good call - I didn't know about this :+1:

grantr:
Thanks for taking this on @chrismwendt!

I feel like there should be a `Percolator` object to represent a percolator, similar to `Warmer`. That would make it slightly easier to implement create/retrieve/delete/exists operations, and avoid putting those in the index method namespace.

Something like this:

```ruby
# Creates a percolator

index.percolator("1").create(:query => { :match => { :author => "pea53" }})
# Retrieves the percolator document

index.percolator("1").get

# Deletes the percolator
index.percolator("1").delete

# Returns true if percolator exists, otherwise false
index.percolator("1").exists?
```

chrismwendt:
That's a great idea, I'll get on it.

lib/elastomer/client/docs.rb
```diff
@@ -261,6 +261,20 @@ def delete_by_query(query, params = nil)
...
+      # index. Type is required.
+      #
+      # Examples
+      #
+      #   index.register_percolator_query 1, { :query => { :match => { :author => "pea53" } } }
+      #   index.docs(type).percolate({ :doc => { :author => "pea53" } })
+      #
+      # Returns the response body as a Hash
+      def percolate(body, params = {})
+        response = client.get '/{index}/{type}/_percolate', update_params(params, :body => body, :action => 'percolator.percolate')
```
grantr: It's possible to percolate an existing document without a body (See the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_percolating_an_existing_document)) so we should support this as well. We could add an optional id element to the URI template like `/{index}/{type}{/id}/_percolate` and allow the body to be omitted.

chrismwendt: :+1: I'll add that.

grantr:
Features we should support:
- [ ] Create/Retrieve/Delete/Exists on percolators
- [ ] Percolate new document
- [ ] Percolate existing document
- [ ] Percolate count new and existing [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_percolator_count_api)
- [ ] Multi percolate [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_multi_percolate_api)

See https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/multi_search.rb for an example of how to implement multi percolate.

chrismwendt:
I copied your task list to the first post so a progress bar appears for this PR on the PR list.

I implemented each of the items :sparkles: so feel free to comment on anything you notice.

lib/elastomer/client/docs.rb
```diff
@@ -261,6 +261,38 @@ def delete_by_query(query, params = nil)
...
+      # include `:id` in the params.
+      #
+      # Examples
+      #
+      #   index.register_percolator_query 1, :query => { :match => { :author => "pea53" } }
+      #   docs.percolate :doc => { :author => "pea53" }
+      #   docs.percolate nil, :id => 3
+      #
+      # Returns the response body as a Hash
+      def percolate(body = nil, params = {})
```
grantr: Since a percolate request with no `body` or `params` is invalid, I think `body` should not have a default value. Same for `percolate_count`.

lib/elastomer/client/index.rb
```diff
@@ -533,6 +533,17 @@ def delete_by_query(query, params = nil)
...
 
+      # Constructs a Percolator with the given id on this index.
+      #
+      # Examples
+      #
+      #   index.percolator "1"
+      #
+      # Returns a Percolator
+      def percolator(id)
+        Percolator.new(@client, @name, id)
```
grantr: Use the `client` and `name` accessors instead of referencing instance variables.

grantr: There's a slight inconsistency here compared to the warmer interface. A `Warmer` object can be constructed two ways:

```ruby
client.warmer('index_name', 'warmer_name')
client.index('index_name').warmer('warmer_name')
```

With the above percolator interface, there's no `client.percolator` option.

Warmers and percolators both require an index scope, so I'm open to removing `Client#warmer` if there's agreement that it's not useful. Either way both objects should have similar interfaces.

lib/elastomer/client/percolator.rb
```diff
@@ -0,0 +1,72 @@
...
+      # client - Elastomer::Client used for HTTP requests to the server
+      # index  - The index name
+      # id     - The _id for the query
+      def initialize(client, index, id)
+        @client = client
+        @index = index
+        @id = id
+      end
+
+      attr_reader :client, :index, :id
```
grantr: `index` should probably be `index_name` to clarify that it's an index name string and not an `Index` object.

lib/elastomer/client/percolator.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      # Create a new Percolator for managing a query.
+      #
+      # client - Elastomer::Client used for HTTP requests to the server
+      # index  - The index name
+      # id     - The _id for the query
+      def initialize(client, index, id)
+        @client = client
+        @index = index
+        @id = id
```
grantr: Use `client.assert_param_presence` to verify that these are valid values. See https://github.com/github/elastomer-client/blob/a19ffa5ab3207ecf162f1a37e70399bad1e31eb7/lib/elastomer/client/warmer.rb#L24-L25

lib/elastomer/client/percolator.rb
```diff
@@ -0,0 +1,72 @@
...
+      # Create a percolator query.
+      #
+      # Examples
+      #
+      #   percolator = $client.index("default-index").percolator "1"
+      #   percolator.create :query => { :match_all => { } }
+      #
+      # Returns the response body as a Hash
+      def create(body, params = {})
+        response = client.put("/#{@index}/.percolator/#{@id}", params.merge(:body => body, :action => 'percolator.create'))
```
grantr: I'd prefer URI templates here, since that's the code path for all other path generation. See https://github.com/github/elastomer-client/blob/a19ffa5ab3207ecf162f1a37e70399bad1e31eb7/lib/elastomer/client/warmer.rb#L42.

lib/elastomer/client/multi_percolate.rb
```diff
@@ -0,0 +1,127 @@
...
+    def multi_percolate(body = nil, params = nil)
+      if block_given?
+        params, body = (body || {}), nil
+        yield mpercolate_obj = MultiPercolate.new(self, params)
+        mpercolate_obj.call
+      else
+        raise 'multi_percolate request body cannot be nil' if body.nil?
+        params ||= {}
+
+        response = self.post '/_mpercolate', params.merge(:body => body)
```
grantr: Should support the other paths as well:
```
/index/type/_mpercolate
/index/_mpercolate
/_mpercolate
```
See https://github.com/github/elastomer-client/blob/a19ffa5ab3207ecf162f1a37e70399bad1e31eb7/lib/elastomer/client/multi_search.rb#L43

lib/elastomer/client/index.rb
chrismwendt: Since elastomer-client aims to provide a one-to-one mapping to the ElasticSearch API endpoints, I think removing `Client#warmer` is the better option.

lib/elastomer/client/multi_percolate.rb
```diff
@@ -0,0 +1,127 @@
...
+      attr_reader :client
+
+      # Add a percolate action to the multi percolate request. This percolate
+      # action will not be executed until the multi_percolate API call is made.
+      #
+      # header - A Hash of the index and type, if not provided in the params
+      # doc    - A Hash of the document
+      #
+      # Returns this MultiPercolate instance.
+      def percolate(header, doc)
```
grantr: To maintain consistency with `MultiSearch` and other document APIs, argument order should be `doc, header`.

lib/elastomer/client/index.rb
grantr: Great, let's do that. I'll open another PR.

lib/elastomer/client/percolator.rb
chrismwendt: Good call - I wasn't sure about the best way to do this.

lib/elastomer/client/multi_percolate.rb
chrismwendt: Good catch - I'll add `Index#multi_percolate` and `Docs#multi_percolate` and those URI template variables so it's consistent with multi search.

lib/elastomer/client/docs.rb
```diff
@@ -261,6 +261,38 @@ def delete_by_query(query, params = nil)
         @client.delete_by_query(query, update_params(params))
       end
 
+      # Matches a provided or existing document to the stored percolator
+      # queries. To match an existing document, pass `nil` as the body and
+      # include `:id` in the params.
+      #
+      # Examples
+      #
+      #   index.register_percolator_query 1, :query => { :match => { :author => "pea53" } }
```
grantr: This method was renamed to `percolator.create` right?

lib/elastomer/client/docs.rb
```diff
@@ -487,6 +519,31 @@ def multi_search( params = {}, &block )
...
+      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_multi_percolate_api
+      #
+      # params - Optional request parameters as a Hash
+      # block  - Passed to a MultiPercolate instance which assembles the
+      #          percolate actions into a single request.
+      #
+      # Examples
+      #
+      #   # block form
+      #   multi_percolate() do |m|
```
grantr: No empty parentheses here

lib/elastomer/client/docs.rb
```diff
@@ -487,6 +519,31 @@ def multi_search( params = {}, &block )
...
+      #   # block form
+      #   multi_percolate do |m|
+      #     m.percolate(:author => "pea53")
+      #     m.count(:author => "grantr")
+      #     ...
+      #   end
+      #
+      # Returns the response body as a Hash
+      def multi_percolate(params = {}, &block)
+        params = { :index => self.name, :type => self.type }.merge params
```
grantr: Use `defaults` here instead

lib/elastomer/client/index.rb
```diff
@@ -501,6 +501,31 @@ def multi_search( params = {}, &block )
...
+      #   # block form
+      #   multi_percolate do |m|
+      #     m.percolate({ :author => "pea53" }, { :type => 'default-type' })
+      #     m.count({ :author => "pea53" }, { :type => 'type2' })
+      #     ...
+      #   end
+      #
+      # Returns the response body as a Hash
+      def multi_percolate(params = {}, &block)
+        params = { :index => self.name }.merge params
```
grantr: Use `defaults` here as well

lib/elastomer/client/multi_percolate.rb
```diff
@@ -0,0 +1,127 @@
...
+      attr_reader :client
+
+      # Add a percolate action to the multi percolate request. This percolate
+      # action will not be executed until the multi_percolate API call is made.
+      #
+      # header - A Hash of the index and type, if not provided in the params
+      # doc    - A Hash of the document
+      #
+      # Returns this MultiPercolate instance.
+      def percolate(doc, header = {})
```
grantr: Is there any reason to allow missing `header`? Seems like it's always required in the request. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-percolate.html#_multi_percolate_api

lib/elastomer/client/multi_percolate.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      # Add a percolate acount action to the multi percolate request. This
+      # percolate count action will not be executed until the multi_percolate
+      # API call is made.
+      #
+      # header - A Hash of the index and type, if not provided in the params
+      # doc    - A Hash of the document
+      #
+      # Returns this MultiPercolate instance.
+      def count(header, doc)
```
grantr: Reverse argument order to match `percolate`

lib/elastomer/client/multi_percolate.rb
chrismwendt: `index` and `type` can be omitted from the header when they're provided in the URL parameters:

```
GET _mpercolate?index=foo&type=bar
{"percolate":{}}
{"doc":{}}
```

It's more of a convenience than anything for `header` to default to `{}`.

lib/elastomer/client/percolator.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      # Create a new Percolator for managing a query.
+      #
+      # client     - Elastomer::Client used for HTTP requests to the server
+      # index_name - The index name
+      # id         - The _id for the query
+      def initialize(client, index_name, id)
+        @client = client
+        @index_name = client.assert_param_presence(index_name, 'index name')
+        @id = id
```
grantr: Is `id` also a required parameter? Should probably `assert_param_presence` on that too.

Bit of historical background, `assert_param_presence` exists because a nil or empty string parameter can cause a command to apply to more things than you expected. See https://github.com/github/elastomer-client/issues/32.

I haven't checked the api, but it's possible that the request `DELETE /index_name/.percolator/` deletes all percolators. In that case we should definitely require a non-blank `id`.

lib/elastomer/client/multi_percolate.rb
grantr: Ah ok, that makes sense. Let's leave it the way it is.

grantr:
Just a few nitpicky things. Excited to get this in!

grantr:
:ship:

chrismwendt:
Thanks for the feedback! :sparkles:

PR #114, opened 2015-09-21
Remove Client#warmer method

grantr:
Since warmers require an index scope, we should be consistent and only provide access from an Index object. This is a breaking change.

/cc @TwP @chrismwendt


chrismwendt:
:+1: :shipit:

PR #115, opened 2015-09-21
Restrict delete_by_query tests to the test index

chrismwendt:
This prevents the `delete_by_query` tests from deleting documents contained in indices other than the test index specifically created by the test.

/cc @grantr

### DIFF
--- a/test/client/docs_test.rb
+++ b/test/client/docs_test.rb
@@ -213,7 +213,7 @@ describe Elastomer::Client::Docs do
     authors = h['docs'].map { |d| d['_source']['author'] }
     assert_equal %w[pea53 grantr], authors
 
-    h = @docs.delete_by_query(:q => "author:grantr")
+    h = @docs.delete_by_query(:q => "author:grantr", :index => @index.name)
     assert_equal(h['_indices'], {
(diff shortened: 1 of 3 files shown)


test/client/index_test.rb
```diff
@@ -366,7 +366,7 @@
     it 'deletes by query' do
       @index.docs('foo').index("foo" => "bar")
       @index.refresh
-      r = @index.delete_by_query(:q => '*')
```
grantr: Shouldn't be necessary here, since there's already an index scope

test/client/docs_test.rb
```diff
@@ -239,14 +239,12 @@
...
+      h = @docs.delete_by_query({
             :query => {
               :filtered => {
                 :query => {:match_all => {}},
-                :filter => {:term => {:author => 'pea53'}}
-              }
-            }
-          )
+                :filter => {:term => {:author => 'pea53'} } } } },
+            :index => @index.name)
```
grantr: No need to add index scope here, since it's already scoped by `@docs`.

chrismwendt:
Closing in favor of #118.

PR #116, opened 2015-09-22
Bump faraday dependency version

shayfrendt:
This updates to the latest Faraday branch. There are a handful of not-backwards-compatible changes in the 0.9.x series of the Faraday gem, but it doesn't seem like any of those changes affect this library.

### DIFF
--- a/elastomer-client.gemspec
+++ b/elastomer-client.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "addressable", "~> 2.3"
-  spec.add_dependency "faraday",     "~> 0.9"
+  spec.add_dependency "faraday",     ">= 0.8"
   spec.add_dependency "multi_json",  "~> 1.7"
   spec.add_dependency "semantic",    "~> 1.3"
 
--- a/elastomer-client.gemspec
+++ b/elastomer-client.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "addressable", "~> 2.3"
-  spec.add_dependency "faraday",     "~> 0.8"
+  spec.add_dependency "faraday",     "~> 0.9"
   spec.add_dependency "multi_json",  "~> 1.7"
   spec.add_dependency "semantic",    "~> 1.3"
 


grantr:
Nice, this will simplify the Faraday dependency management. If we're going to support both0.8 and 0.9, it would be nice to have ci that tests both gems.

shayfrendt:
>Nice, this will simplify the Faraday dependency management.

:cool: :sunflower: :sunglasses: 

> it would be nice to have ci that tests both gems.

@grantr You mean like including the 0.8 and 0.9 branches of Faraday in the CI build matrix for elastomer-client? Haven't done that for a gem in a while, but it looks like there are a few things out there meant to help with that goal and the [`.travis.yml` file](https://github.com/github/elastomer-client/blob/master/.travis.yml), like https://github.com/thoughtbot/appraisal.

grantr:
Appraisal looks nice. Do we use it anywhere?

shayfrendt:
Doesn't look like we're using it, but on an unrelated/related note we recognize the Appraisals filename via linguist: https://github.com/github/linguist/pull/764/files

shayfrendt:
Here are some reference implementations for https://github.com/thoughtbot/appraisal as well:
- https://github.com/thoughtbot/factory_girl/blob/master/Appraisals
- https://github.com/johnbintz/rack-livereload/blob/master/Appraisals
- https://github.com/zuorasc/zuora/commit/8ce0f4a6754a2af0367f0b3792eecea11816d1a0

shayfrendt:
@grantr From a mergeability standpoint, do you think we need to get all of the appraisal stuff set up now, or can that be added to the gem maintenance backlog so we can go ahead and get this changed merged down?

grantr:
@shayfrendt I don't think appraisal is necessary to get this merged as long as you're confident it works in your app.

shayfrendt:
:cool: I'll get this merged shortly.

TwP:
What is the intent behind this change? Apologies for jumping in so late on the conversation here.

With a version dependency of `~> 0.8`, we can use any Faraday version up to but excluding a 1.0 release. Changing the version dependency to `=> 0.8` means that `elastomer-client` will try and use the next major release of Faraday when it comes out. Major releases introduce backwards incompatibilities (according to [semantic versioning](http://semver.org) at least).

My vote is to revert this change as it could break `elastomer-client` when the next major release of Faraday happens.

shayfrendt:
Ah yeah, from a maintenance perspective it would be not amazing to auto-jump to the next major release. I'll revert this.

PR #118, opened 2015-09-22
Scope the delete by query tests to a single index

TwP:
Running the delete by query tests locally resulted in all documents in my local ES installation being deleted. The delete by query tests were scoped at the `$client` level, not at the `@index` level. So the scan / bulk delete operations iterated over all the indices in my development ES step :disappointed: 

These changes scope the tests to just the test index :smile:

/cc @chrismwendt @grantr


grantr:
:+1:

chrismwendt:
This solves the issue in a simpler way, thanks :bow:

Issue #117, opened 2015-09-22
Exclude source retrieval in delete_by_query


TwP:
The `delete_by_query` call should exclude the document source from being retrieved when performing the scan operation. We only need the `index`, `type`, and `id` of the documents in order to delete them. Excluding source retrieval will reduce load on the search cluster, and it will reduce the size of our HTTP packets.

```ruby
@query = query.dup
@query.delete :_source if @query.key? :_source
@query.delete "_source" if @query.key? "_source"
@query[:_source] = false 
```

Example implementation to put into the `DeleteByQuery` initializer.

/cc @chrismwendt @grantr

grantr:
Definitely :+1: although we should check to make sure all of our deployed versions support this.

TwP:
Agreed - I was attempting to dig through the Elasticsearch docs to find when / where the `_source` field was introduced. It was formerly the `fields` keyword in the query document. I'll do a little more digging.

chrismwendt:
According to https://www.elastic.co/guide/en/elasticsearch/reference/1.3/search-request-source-filtering.html#search-request-source-filtering, the `_source` field was added in ES 1.0.0.Beta1.

chrismwendt:
Done in #119.

PR #119, opened 2015-09-22
Omit source in delete_by_query searches

chrismwendt:
This prevents the unused `_source` field of search results from being sent over the network.

Solves #117.

/cc @TwP @grantr


TwP:
:+1: I just learned that you can pass `_source` via the parameters and it will override any `_source` fields in the query body:

```bash
curl -XPOST 'localhost:19200/users/user/_search?_source=false' -d '{
  "query": {"match_all":{}},
  "_source": "*email"
}'
```

PR #121, opened 2015-09-27
Use VCR to make tests deterministic

chrismwendt:
This PR introduces VCR for making tests deterministic:

- [x] hot_threads

@TwP @grantr: Have you seen other nondeterministic tests?

### DIFF
--- a/elastomer-client.gemspec
+++ b/elastomer-client.gemspec
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.5"
   spec.add_development_dependency "activesupport", ">= 3.0"
   spec.add_development_dependency "minitest","~> 4.7"
+  spec.add_development_dependency "vcr","~> 2.9"
   spec.add_development_dependency "webmock","~> 1.21"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "overcommit"
--- a/test/client/nodes_test.rb
+++ b/test/client/nodes_test.rb
@@ -39,9 +39,11 @@ describe Elastomer::Client::Nodes do
   end
 
   it "gets the hot threads for the node(s)" do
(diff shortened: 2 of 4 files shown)


chrismwendt:
One of the index tests failed due to a timeout on line https://github.com/github/elastomer-client/blob/vcr/test/client/index_test.rb#L112

```
Elastomer::Client::Index#test_0004_updates index settings:
Elastomer::Client::TimeoutError: Net::ReadTimeout :: POST /elastomer-index-test
call stack...
```

If such an innocent test can fail, then I wonder if all of the tests should be VCRed.

TwP:
My first thought is to use VCR sparingly. The reason being we want to find discrepancies between our client implementation and the Elasticsearch API quickly. Having VCR in the mix might hide errors.

One solution for that concern is to script cassette creation. Hmmm ... can we have cassettes for each version of ES that we support? That would enable us to test against multiple version of ES without requiring multiple versions of ES to be running. But that might be far more complex than is worth doing.

grantr:
I'd prefer to use real requests as much as possible for the reasons @TwP mentioned. We want to make sure elastomer client actually works with the version of ES under test. The tests aren't just testing elastomer-client, they're also testing whether ES is fulfilling the contract that elastomer-client expects.

If we're getting timeouts, maybe we could increase the read timeout option for the test client.

chrismwendt:
Good points - it sounds like testing against multiple versions of ES is something we want to tackle before using VCR. I'll just note here that this can be done via Travis with environment variables and `before_install` commands: https://github.com/bitemyapp/bloodhound/blob/master/.travis.yml

For now, in a new PR I'll increase the timeout as @grantr suggested, and change the regex on the `hot_threads` test.

chrismwendt:
Fixed the `hot_threads` test by changing the regex in #122.

PR #122, opened 2015-09-28
Make tests more stable

### LINKED ISSUE #120: Make the hot_threads test deterministic
The hot_threads test fails nondeterministically, which causes false negatives during builds.

```
Elastomer::Client::Nodes#test_0005_gets the hot threads for the node(s) [/var/lib/jenkins/workspace/elastomer-client/test/client/nodes_test.rb:44]:
Expected /cpu usage by thread/ to match "
```

chrismwendt:
This makes the `hot_threads` test deterministic and increases the stability of other tests by increasing the `read_timeout`.

Fixes #120.

/cc @TwP @grantr

### DIFF
--- a/test/client/nodes_test.rb
+++ b/test/client/nodes_test.rb
@@ -41,7 +41,8 @@ describe Elastomer::Client::Nodes do
   it "gets the hot threads for the node(s)" do
     str = $client.nodes.hot_threads :read_timeout => 2
     assert_instance_of String, str
-    assert_match %r/:::/, str
+    assert !str.nil?, "expected response to not be nil"
+    assert !str.empty?, "expected response to not be empty"
   end
 
   it "can be scoped to a single node" do
--- a/test/client/nodes_test.rb
+++ b/test/client/nodes_test.rb
@@ -41,7 +41,7 @@ describe Elastomer::Client::Nodes do
   it "gets the hot threads for the node(s)" do
     str = $client.nodes.hot_threads :read_timeout => 2
     assert_instance_of String, str
-    assert_match %r/Hot threads/, str
+    assert_match %r/:::/, str
   end
 
   it "can be scoped to a single node" do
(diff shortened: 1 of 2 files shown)


test/client/nodes_test.rb
```diff
@@ -41,7 +41,7 @@
   it "gets the hot threads for the node(s)" do
     str = $client.nodes.hot_threads :read_timeout => 2
     assert_instance_of String, str
-    assert_match %r/cpu usage by thread/, str
+    assert_match %r/:::/, str
```
TwP: Trying this out locally, and I noticed that my `hot_threads` call returned zero threads. Elasticsearch introduced the `ignore_idle_threads` parameter which defaults to `true`. So if the cluster is largely idle, there will be no hot threads to output. Setting the `ignore_idle_threads` param to `false` gives some output text to match against.

This param was added in Dec 2014 https://github.com/elastic/elasticsearch/commit/242e631e95a1b9240d3d84b16912c69bc63f031b

chrismwendt: It appears that that commit targets ES 2.0, so we would need to bump from 1.4.0 to 2.0 for Travis, right? I'm all for making the test as accurate as possible. Silly non-JSON format...

TwP: That commit is on the master branch of elastic/elasticsearch. So it is included in the current ES release. Not sure what the 2.0.0 beta tags are about.

grantr: For hot_threads specifically, I'm not too worried about test correctness. It's an admin endpoint that's only useful for human viewers. I'd be willing to drop its tests if they're causing problems.

chrismwendt: @TwP: Yeah that's weird, and the docs for 1.4.0 don't have that param: https://www.elastic.co/guide/en/elasticsearch/reference/1.4/cluster-nodes-hot-threads.html#cluster-nodes-hot-threads

How about just a check for HTTP OK?

TwP: :+1: to HTTP OK **and** checking for a non-empty response body. We really don't care what is returned just so long as *something* is returned.

grantr: :+1:

chrismwendt: `Client#hot_threads` returns just the body, so I checked that and not the HTTP status.

PR #126, opened 2016-01-15
Update documentation

wags:
Some housekeeping-related updates to the documentation. Standardize on `Elasticsearch` rather than `ElasticSearch`, update URLs to point to `https://www.elastic.co` rather than `http://www.elasticsearch.org` and fix a few typos.


chrismwendt:
Thanks for the updates, @wags! Those links had certainly begun to age. Looks good to me :+1:

PR #130, opened 2016-02-26
Suggest endpoint

TwP:
This PR adds support for the [suggesters API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html) in Elasticsearch.

/cc @chrismwendt @grantr


grantr:
:+1:

PR #131, opened 2016-05-05
Default bulk streaming request_size to 10MB

chrismwendt:
I keep getting bit by a hanging process, only to find that I forgot to set the `request_size` when streaming bulk operations. This sets a default `request_size` of 10MB.

cc @github/search

### DIFF
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -174,7 +174,7 @@ module Elastomer
         @actions = []
         @current_request_size = 0
         @current_action_count = 0
-        self.request_size = params.delete(:request_size)
+        self.request_size = params.delete(:request_size) || 10 * 1024 * 1024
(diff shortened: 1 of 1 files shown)


lib/elastomer/client/bulk.rb
```diff
@@ -174,7 +174,7 @@ def initialize( client, params = {} )
         @actions = []
         @current_request_size = 0
         @current_action_count = 0
-        self.request_size = params.delete(:request_size)
+        self.request_size = params.delete(:request_size) || 10 * 1024 * 1024
```
TwP: I would recommend changing this magic number to a constant and then use the constant here.

```ruby
class Bulk
  DEFAULT_REQUEST_SIZE = 10 * 2**20  # 10 MB
  ...
end
```

TwP:
This is a good change to make :+1: Thank you :bow:

PR #135, opened 2016-09-01
Fix delete by query with required routing

chrismwendt:
This fixes delete by query when routing is required.

In ES 1.x, `_routing` is not required when deleting a document, and `_routing` is not returned in query results.

In ES 2.x, `_routing` is required when deleting a document if the mapping requires `_routing`, and `_routing` is returned in query results.


TwP:
Sweet! Thank you for starting with a failing test. Good catch finding this issue :sparkles:

PR #136, opened 2016-09-02
Enforce request size limits

TwP:
The goal of this PR is to add some default request size limits to elastomer client. Currently with our `bulk` indexing layer, it is possible to send very very large requests to Elasticsearch - multi-gigabyte requests.

This PR implements a `limit_size` request middleware that will raise a `RequestSizeError` if the request body exceeds a a `max_request_size`. This limit defaults to 250 MB, but it is configurable when the `Elastomer::Client` is first created.

This PR also reworks the criteria determining when an actual bulk request is set. As bulk request actions are accumulated, we now perform a pre-check on the request size to ensure that a large document does not push any single bulk request above this request size. Any single document that is larger than the request size will also raise a `RequestSizeError` exception.

/cc @chrismwendt @grantr

### DIFF
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -158,7 +158,7 @@ module Elastomer
     # immediately.
     #
     class Bulk
-      DEFAULT_REQUEST_SIZE = 10 * 2**20 # 10 MB
+      DEFAULT_REQUEST_SIZE = 10 * 2**20  # 10 MB
 
       # Create a new bulk client for handling some of the details of
       # accumulating documents to index and then formatting them properly for
@@ -194,7 +194,10 @@ module Elastomer
         if value.nil?
           @request_size = nil
         else
-          @request_size = value.to_i <= 0 ? nil : value.to_i
+          value = value.to_i
+          value = nil if value <= 0
+          value = client.max_request_size if value > client.max_request_size
+          @request_size = value
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/elastomer/client/bulk.rb
```diff
@@ -158,7 +158,7 @@ def bulk_stream_items(ops, params = {})
     # immediately.
     #
     class Bulk
-      DEFAULT_REQUEST_SIZE = 10 * 2**20 # 10 MB
+      DEFAULT_REQUEST_SIZE = 10 * 2**20  # 10 MB
```
chrismwendt: Eep, extra space?

TwP:
`git` needs a spellchecker on branch names :disappointed:

lib/elastomer/client/bulk.rb
```diff
@@ -370,26 +373,55 @@ def convert_special_keys(params)
...
       end
 
-    end  # Bulk
-  end  # Client
-end  # Elastomer
+      # Internal: Raises a RequestSizeError if the given size is larger than
+      # the configured requst size or the client.max_request_size
+      def check_action_size!( size )
+        return unless (request_size && size > request_size) || (size > client.max_request_size)
+        raise RequestSizeError, "Bulk action of size `#{size}` is too large"
```
chrismwendt: Should we emit the index, type, and doc ID here to make debugging easier?

TwP: I thought about that, and that type of error logging should happen at a higher level. An application would catch this exception and then log pertinent information from the application context.

PR #137, opened 2016-09-02
Exception wrapping

TwP:
This PR is aimed at adding some exception wrappers around common error responses from Elasticsearch. The problem is that in some of our error handling code we have lines like this:
    
```ruby
rescue StandardError => err
  if err.message =~ %r/IndexMissingException|index_not_found_exception/
    # do someting here
  end
```
    
A better approach is to rescue a named exception and not worry about using a regular expression on the error message to figure out what the problem actually is.

/cc @chrismwendt @grantr

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -313,7 +313,8 @@ module Elastomer
         elsif error.is_a?(String)
           case error
           when %r/IndexMissingException/; raise IndexNotFoundError, response
-          when %/QueryParsingException/; raise QueryParsingError, response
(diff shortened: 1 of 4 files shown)


chrismwendt:
Good call to do this once in elastomer-client rather than all over the place in downstream libraries and apps :sparkles:

TwP:
The plan is to roll out these exception wrappers one by one as we have need of them. Currently I'm doing this type of check in a few places, and I'd rather rescue this `IndexNotFoundError`. Can you think of any other error checking that we are doing using regular expression matching?

PR #139, opened 2016-11-22
Freeze the body to prevent excon from changing its encoding

### LINKED ISSUE #138: StringIO can change `body` String encodings
We ran into an issue where the encoding of a `body` string passed to a GET request was being changed by the underlying [Excon](https://github.com/excon/excon) Faraday adapter. The encoding was changed from UTF-8 to ASCII, and it caused exceptions further up the stack when we attempted to render analytics information about individual Elasticsearch requests.

Excon wraps the `body` of a request with a `StringIO` instance. This allows Excon to [read the body in chunks](https://github.com/excon/excon/blob/master/lib/excon/connection.rb#L166) and send those chunks in the HTTP request. Excon calls `StringIO#binmode` to force binary encoding such that the chunks are not affected by multibyte UTF-8 characters. The problem is that the encoding of the underlying String is also changed by this `binmode` call.

We should modify the `Elastomer::Client#extract_body` method to return frozen Strings. This will prevent this type of error, and others like it, from occurring in the future.

chrismwendt:
This prevents excon from [changing the encoding of the body](https://github.com/excon/excon/blob/v0.32.1/lib/excon/connection.rb#L183), causing downstream encoding compatibility errors.

Fixes #138

/cc @github/search

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -162,6 +162,9 @@ module Elastomer
       body = extract_body params
       path = expand_path path, params
 
+      # prevent excon from changing the encoding
+      body.freeze
+
       instrument(path, body, params) do
         begin
           response =
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -162,7 +162,7 @@ module Elastomer
       body = extract_body params
       path = expand_path path, params
 
-      # prevent excon from changing the encoding
+      # Prevent excon from changing the encoding (see https://github.com/github/elastomer-client/issues/138)
       body.freeze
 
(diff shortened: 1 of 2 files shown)


lib/elastomer/client.rb
```diff
@@ -162,6 +162,9 @@ def request( method, path, params )
       body = extract_body params
       path = expand_path path, params
 
+      # Prevent excon from changing the encoding (see https://github.com/github/elastomer-client/issues/138)
+      body.freeze
```
chrismwendt: Good catch, moved to `extract_body` in 7474cf3

PR #140, opened 2017-11-08
Update instructions for running tests and fix Janky build

look:
There is no `testrb` in this repository so `script/test` does not work as
advertised. `rake test` is the current thing to use.


TwP:
:+1: to merging this in and then fixing the Travis-CI builds in a separate PR

look:
Sounds good. Debugging a failing Travis build usually takes a couple of commits.

PR #148, opened 2017-11-17
Removing ES 1.X code paths

TwP:
This PR removes all the Elasticsearch 1.X (and earlier) code paths from the client gem. We will no longer be supporting any versions of Elasticsearch prior to Elasticsearch 2.X.

/cc @look && @elireisman

### DIFF
--- a/test/client/index_test.rb
+++ b/test/client/index_test.rb
@@ -166,37 +166,6 @@ describe Elastomer::Client::Index do
     assert_property_exists @index.mapping[@name], "mux_mool", "song"
   end
 
-  it "deletes document mappings" do
-    @index.create(
-      :mappings => {
-        :doco => {
-          :_source => { :enabled => false },
-          :_all    => { :enabled => false },
-          :properties => {:title  => { :type => "string", :analyzer => "standard" }}
-        }
-      }
-    )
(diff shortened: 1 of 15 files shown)


test/client/index_test.rb
```diff
@@ -270,15 +222,11 @@
     before do
       suggest = {
         :type            => "completion",
-        :index_analyzer  => "simple",
+        :analyzer        => "simple",
```
look: since you did the other hashes, should you do this one too?

TwP: Eventually ... I realized I was trying to do "all the things" in a single pull request. I did this one here because of the change lower down where I removed the `es_version_2_x?` check.

test/client/nodes_test.rb
```diff
@@ -16,26 +16,24 @@
...
-      assert !node.key?("network"), "expected network info to be absent"
-    end
-
-    it "filters node stats" do
-      h = $client.nodes.stats(:stats => "http")
-      node = h["nodes"].values.first
-      assert node.key?("http"), "expected http stats to be present"
-      assert !node.key?("indices"), "expected indices stats to be absent"
-    end
+  it "fitlers node info" do
```
elireisman: unimportant nitpick, but `s/fitlers/filters` ✏️

PR #150, opened 2017-11-17
Add minitest-fail-fast

look:
This will allow halting tests on the first failure, which should make updating
individual failing tests on ES 5.6 a little easier and faster.

To run, use:

    TESTOPTS="--fail-fast" bin/rake

In order to do this, minitest had to be upgraded to 5+. I verified that all the
tests continue to work on this version.

cc: @elireisman


elireisman:
@look I'll await @TwP's opinion but LGTM! Not Rubylicious enough yet to know if checking in the Gemfile.lock is more or less "correct" but seems reasonable as well since tracking changes in it as we pick up version updates can't be a bad thing? Any thoughts about whys/hows here would be educational for me for sure if you have some :)

look:
@elireisman checking in the `Gemfile.lock` is usually recommended because that is what `bundle install` uses to install the gems. If you don't check in the file, it does a fresh dependency resolution and you'll end up with a (possibly dramatically) different environment.

However, it's not recommended for gems, which is what this is. So maybe a better solution to this would be to update the CI script to run `bundle update` before executing. What happened here is that without this file OR `bundle update` our CI environment wasn't able to install the new gem I added. Travis CI had no problem because it started fresh.

[This is a good article about it](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/):

> When developing a gem, use the gemspec method in your Gemfile to avoid duplication. In general, a gem's Gemfile should contain the Rubygems source and a single gemspec line. Do not check your Gemfile.lock into version control, since it enforces precision that does not exist in the gem command, which is used to install gems in practice. Even if the precision could be enforced, you wouldn't want it, since it would prevent people from using your library with versions of its dependencies that are different from the ones you used to develop the gem.
>
> When developing an app, check in your Gemfile.lock, since you will use the bundler tool across all machines, and the precision enforced by bundler is extremely desirable for applications.

TwP:
> So maybe a better solution to this would be to update the CI script to run `bundle update` before executing.

Give that a shot ?? It sounds like a reasonable thing to do given that this is a gem.

look:
Good call. I talked to developer-experience and confirmed that Janky builds pollute their filesystem.

The ultimate solution to this is Dockerizing the build, which I don't really want to do right now, but seems useful to know about.

So for now, I've updated the build script to delete the lockfile before installing gems and that worked. I like this better.

elireisman:
@look thanks for the article/explain, makes sense!

PR #152, opened 2017-11-21
Use rm -f on Gemfile.lock

look:
If the Gemfile.lock WASN'T there, the build would fail. :(
This will ignore the error and allow the build to proceed.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -19,7 +19,7 @@ if [ -d /usr/share/rbenv/shims ]; then
     export PATH=/usr/share/rbenv/shims:$PATH
 fi
 ruby -v
-rm Gemfile.lock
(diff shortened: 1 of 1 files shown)


script/cibuild
```diff
@@ -19,7 +19,7 @@ if [ -d /usr/share/rbenv/shims ]; then
     export PATH=/usr/share/rbenv/shims:$PATH
 fi
 ruby -v
-rm Gemfile.lock
+rm -f Gemfile.lock
```
elireisman: This LGTM! Could also put the `rm Gemfile.lock` between `set +e ... set -e` block if missing file is the issue causing CI to stop but this seems safe, I like it! :shipit:

PR #155, opened 2017-11-30
Use `payloads` conditionally for 5.x compatibility

juruen:
Use `payloads` conditionally as 5.x drops support for index-time payloads.

I did it this way to keep the old behavior but maybe we could just remove `payloads` altogether for both versions?

### DIFF
--- a/test/client/index_test.rb
+++ b/test/client/index_test.rb
@@ -224,8 +224,11 @@ describe Elastomer::Client::Index do
         :type            => "completion",
         :analyzer        => "simple",
         :search_analyzer => "simple",
-        :payloads        => false
       }
+
+      # COMPATIBILITY
+      # ES 5.x drops support for index-time payloads
+      suggest[:payloads] = false if es_version_2_x?
 
       @index.create(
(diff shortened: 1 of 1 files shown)


test/client/index_test.rb
```diff
@@ -224,9 +224,12 @@
         :type            => "completion",
         :analyzer        => "simple",
         :search_analyzer => "simple",
-        :payloads        => false
       }
 
+      # COMPATIBILITY
+      # ES 5.x drops support for index-time payloads
+      suggest[:payloads] = false if es_version_2_x?
+
```
TwP: I very much like the `COMPATIBILITY` tag here. Making a note for myself to write up the tags we are using to annotate the codebase along with what those tags should be used for.

TwP:
We can merge this pull request after we have cut the official 2.3.0 release of the gem.

look:
2.3.0 has been tagged and released: https://rubygems.org/gems/elastomer-client/versions/2.3.0

This should be good to merge.

PR #156, opened 2017-11-30
rename `Index#optimize` to `Index#forcemerge`

juruen:
This was deprecated in favor of `forcemerge` in 2.1:

https://www.elastic.co/guide/en/elasticsearch/reference/2.1/indices-optimize.html

A few options if we would want to break the API:

1. Use an `alias_method` for `optimize` that would use `forcemerge`. However, this wouldn't work with ES 2.0.x, i.e: 2.0, 2.0.1 and 2.0.2.
2. Use approach (1) but detect whether we are targeting ES 2.0.x to use `_optimize` instead.
3. Support both  `optimize` and `forcemerge` at the same time.

Update: we ended up taking  approach (1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 2.3.0 (2017-11-29)
 - Remove Elasticsearch 1.x and earlier code paths
 - Fix CI and configure an Elasticsearch 5.6 build
-- BREAKING: Rename `Index#optimize` to `Index#forcemerge`
 
 ## 2.2.0 (2017-04-29)
 - Added a `clear_scroll` API
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.3.0 (2017-11-29)
 - Remove Elasticsearch 1.x and earlier code paths
 - Fix CI and configure an Elasticsearch 5.6 build
+- BREAKING: Rename `Index#optimize` to `Index#forcemerge`
 
 ## 2.2.0 (2017-04-29)
 - Added a `clear_scroll` API
--- a/lib/elastomer/client/index.rb
+++ b/lib/elastomer/client/index.rb
@@ -261,17 +261,17 @@ module Elastomer
         response.body
       end
 
-      # Optimize one or more indices. Optimizing an index allows for faster
-      # search operations but can be resource intensive.
+      # Force merge one or more indices. Force merging an index allows to
+      # reduce the number of segments but can be resource intensive.
       #
       # params - Parameters Hash
-      #   :index - set to "_all" to optimize all indices
+      #   :index - set to "_all" to force merge all indices
       #
-      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-optimize.html
+      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html
       #
       # Returns the response body as a Hash
-      def optimize( params = {} )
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.3.0 (2017-11-29)
 - Remove Elasticsearch 1.x and earlier code paths
 - Fix CI and configure an Elasticsearch 5.6 build
+- BREAKING: Rename `Index#optimize` to `Index#forcemerge`
```
TwP: This change will not be included in the 2.3.0 release. The goal of the 2.3.0 release is to remove the ES 1.X code paths and make the library compatible **only** with ES 2.X.

This change will go out in the next release.

TwP: This change will not be included in the 2.3.0 release. The goal of the 2.3.0 release is to remove the ES 1.X code paths and make the library compatible **only** with ES 2.X.

This change will go out in the next release.

lib/elastomer/client/index.rb
```diff
@@ -261,17 +261,17 @@ def flush( params = {} )
...
       #
-      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-optimize.html
+      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html
       #
       # Returns the response body as a Hash
-      def optimize( params = {} )
-        response = client.post "{/index}/_optimize", update_params(params, :action => "index.optimize")
+      def forcemerge( params = {} )
+        response = client.post "{/index}/_forcemerge", update_params(params, :action => "index.forcemerge")
         response.body
```
TwP: Could you please also add an `:optimize` alias for this `:forcemerge` function such that we can maintain backwards compatibility? Please add a "DEPRECATED:" comment above the alias so we know to remove it in the future.

```ruby
# DEPRECATED:  ES 5.X has removed the `/_optimize` endpoint.
alias :optimize :forcemerge
```

juruen: @TwP sure, just one more thing related to this that I added in the  PR  description. If we just add an alias, I believe this will fail with ES 2.0.x because the API renaming happened in 2.1.

TwP: I'm fine with breaking ES 2.0. That version of Elasticsearch has been End-of-Lifed by Elastic, so I'm not interested in maintaining support there.

And yes, we are loosey goosey with our version support, and tightening up on version is another project. With parameter validation coming in ES 6.X we will have to be more strict anyways.

juruen: Understood, thanks for the clarification! :)

juruen: That's now changed as discussed. Also, I try to exercise the alias in the test.

look:
@juruen do you mind updating the PR title now that this is no longer a breaking change? Thanks!

juruen:
@look title changed, thanks!

test/client/index_test.rb
```diff
@@ -270,8 +270,12 @@
       assert_equal 0, response["_shards"]["failed"]
     end
 
-    it "optimizes" do
-      response = @index.optimize
+    it "force merges" do
+      if es_version_5_x?
```
look: Since ES 2 has forcemerge and Elastomer is aliasing it to `optimize` as a courtesy for previous users of this library, I don't think this test needs to be conditionalized -- it should work on either ES version with either method call.

You can test both versions if you like.

juruen: That makes sense!

juruen:
Thank you 🎉 !

PR #160, opened 2017-11-30
New assertions to document 2.x/5.x alias behavior

look:
In Elasticsearch 2.x, requesting an alias that does not exist will return an
empty result. But in Elasticsearch 5.x, it will raise an error. This applies to
both full matches and wildcards.

This documents the difference, it doesn't do anything to paper it over.

cc: @juruen @elireisman


TwP:
Weird decision on Elastic's part to return an error. Is it returning a 404 now instead of a 200?

PR #161, opened 2017-11-30
Deprecate warmers

look:
Warmers have been deprecated as of ES 2.3 and removed from Elasticsearch as of
5.0:

https://www.elastic.co/guide/en/elasticsearch/reference/2.3/indices-warmers.html
https://www.elastic.co/guide/en/elasticsearch/reference/5.0/indices-warmers.html

I have marked Elastomer::Client::Warmer as DEPRECATED and skipped the tests when
running under ES 5.x.

### DIFF
--- a/lib/elastomer/client/warmer.rb
+++ b/lib/elastomer/client/warmer.rb
@@ -1,6 +1,8 @@
 module Elastomer
   class Client
 
+    # DEPRECATED: Warmers have been removed from Elasticsearch as of 5.0.
+    # See https://www.elastic.co/guide/en/elasticsearch/reference/5.0/indices-warmers.html
     class Warmer
 
       # Create a new Warmer helper for making warmer API requests.
@@ -66,6 +68,7 @@ module Elastomer
       # See https://github.com/elasticsearch/elasticsearch/issues/5155
       def exists?
         response = get
+
         response != {}
       rescue Elastomer::Client::Error => exception
         if exception.message =~ /IndexWarmerMissingException/
--- a/lib/elastomer/client/warmer.rb
+++ b/lib/elastomer/client/warmer.rb
@@ -11,6 +11,10 @@ module Elastomer
       # index_name - The name of the index as a String
       # name       - The name of the warmer as a String
       def initialize(client, index_name, name)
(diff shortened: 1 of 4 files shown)


elireisman:
Ah crap I think I was doing the same thing (along with a couple of the other test fixes) - this LGTM!

look:
We should consider checking the Elasticsearch version and raising errors for all methods in the `Warmer` class.

TwP:
Checking the Elasticsearch version should be very doable but kinda tedious in places. The tedium can be dealt with though. The nice thing is we memoize the version in the `Client`, so the check will not be heavyweight at all.

https://github.com/github/elastomer-client/blob/f1a1c72c5a90c852a729b5f87f9744ea6725a695/lib/elastomer/client.rb#L54-L57

We reuse the `Client` connections, so that version will stick around.

look:
After more discussion, @TwP and I agreed we should raise errors if you try to use the Warmer class in ES 5. The reason is that you get weird error messages now:

```
Elastomer::Client::RequestError: {"root_cause"=>[{"type"=>"invalid_type_name_exception", "reason"=>"Document mapping type name can't start with '_', found: [_warmer]"}], "type"=>"invalid_type_name_exception", "reason"=>"Document mapping type name can't start with '_', found: [_warmer]"}
```

It would be better if this failed fast.

elireisman:
I like that idea (raising useful error) I was actually adding a SupportedVersionsError for some of the version shims for percolator API for this, but will adapt that before PR to whatever matches up here. This LGTM!

lib/elastomer/client/warmer.rb
```diff
@@ -66,6 +72,7 @@ def get(params = {})
       # See https://github.com/elasticsearch/elasticsearch/issues/5155
       def exists?
         response = get
+
```
TwP: Is this newline intentional?

TwP:
Changes look good to me :rocket:

lib/elastomer/client/warmer.rb
look: I can remove it. I originally had some code here before I realized warmers were Just Gone in 5.

PR #162, opened 2017-12-01
pass scroll_id as a json object

juruen:
Passing the scroll_id directly in the request body
without being wrapped in a json object makes
ES 5.X unable to parse it.

It seems that previous ES versions supported both a json object and just the id.

cc @look and @elireisman


look:
LGTM.

One of the builds was failing but it was an ES start up problem. I restarted it and it passed now. We may need to increase how long we wait for ES to come up in the build container.

look:
This has been open for a while so I'm going to merge it now. Let me know if you'd prefer I don't merge your PRs for you. Thanks! :shipit:

PR #163, opened 2017-12-01
Implement scan through scroll plus sort by doc (fixes delete_by_query in 5.x)

juruen:
This is my first attempt at fixing the deprecated `scan` search_type in 5.x.

First of all, this depends on another super simple PR https://github.com/github/elastomer-client/pull/162 that fixes how `scroll_id` is passed.

The recommended way to replace the `scan` functionality according to the ES doc https://www.elastic.co/guide/en/elasticsearch/reference/2.1/breaking_21_search_changes.html#_literal_search_type_scan_literal_deprecated is to do the following:

>search_type=scan deprecatededit
The scan search type has been deprecated. All benefits from this search type can now be achieved by doing a scroll request that sorts documents in _doc order, for instance:
```
GET /my_index/_search?scroll=2m
{
  "sort": [
    "_doc"
  ]
}
```

We do so in the query that we send to ES.

The other thing that we address in this PR is the fact that some internal parameters of the library were being leaked to the URI that is used for the request. For instance, we would hit the `_search` end point with `action_count` or the `_bulk` end point with `q:name=foo`. Previous versions of ES were more lenient regarding this, but ES 5.x is strict about it. My current approach is pretty naive: try to filter the parameters accordingly.

As it stands right now, the `delete_by_query` test passes in both versions.

cc @look and @elireisman

### DIFF
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -144,9 +144,7 @@ module Elastomer
 
       # Remove parameters that are not valid for the _bulk endpoint
       def bulk_params
-        params = @params.clone
-        params.delete(:q) if params.has_key?(:q)
-        params
+        @params.clone.tap { |p| p.delete(:q) }
       end
 
     end  # DeleteByQuery
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -136,13 +136,13 @@ module Elastomer
         @response_stats
       end
 
-      # Remove parameters that are not valid for the _search endpoint
+      # Internal: Remove parameters that are not valid for the _search endpoint
       def search_params
         params = @params.merge(:_source => false)
         params.select {|p, _| SEARCH_PARAMETERS.include? p}
       end
 
-      # Remove parameters that are not valid for the _bulk endpoint
+      # Internal: Remove parameters that are not valid for the _bulk endpoint
       def bulk_params
         @params.clone.tap { |p| p.delete(:q) }
       end
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -1,5 +1,3 @@
-require 'active_support/core_ext/hash'
-
 module Elastomer
   class Client
 
@@ -146,7 +144,7 @@ module Elastomer
 
       # Remove parameters that are not valid for the _bulk endpoint
       def bulk_params
-        @params.except :q
+        @params.clone.tap { |p| p.delete(:q) }
       end
 
     end  # DeleteByQuery
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -48,6 +48,27 @@ module Elastomer
       DeleteByQuery.new(self, query, params).execute
     end
 
+    SEARCH_PARAMETERS = [
+      :q,
+      :df,
+      :analyzer,
+      :analyze_wildcard,
+      :batched_reduce_size,
+      :default_operator,
+      :lenient,
+      :explain,
+      :_source,
+      :stored_fields,
+      :sort,
+      :track_scores,
+      :timeout,
+      :terminate_after,
+      :from,
+      :size,
+      :search_type,
+      :scroll
+    ].to_set.freeze
+
     class DeleteByQuery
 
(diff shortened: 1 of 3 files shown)


lib/elastomer/client/delete_by_query.rb
```diff
@@ -48,6 +48,27 @@ def delete_by_query(query, params = {})
...
+      :sort,
+      :track_scores,
+      :timeout,
+      :terminate_after,
+      :from,
+      :size,
+      :search_type,
+      :scroll
+    ].to_set.freeze
+
```
juruen: I took all the valid search parameters from the doc. I'm not entirely sure we need to add all of them though.

lib/elastomer/client/scroller.rb
```diff
@@ -45,7 +45,13 @@ def scroll( query, opts = {} )
     #
     # Returns a new Scroller instance
     def scan( query, opts = {} )
-      opts = opts.merge(:search_type => "scan")
+      if query.nil?
+        query = {}
+      elsif query.is_a? String
+        query = MultiJson.load(query)
+      end
+      query = query.merge(:sort => [:_doc])
```
look: One thing I was thinking about with this change was to fail fast if the passed in query had a sort. What do you think? 

The nice thing about that is if you were expecting sorting, it will complain loudly.

juruen: That makes a lot of sense. I understand you correctly you are proposing to raise an exception right away if the query contains a `:sort` already.

lib/elastomer/client/delete_by_query.rb
```diff
@@ -105,16 +126,27 @@ def accumulate(item)
...
 
+      # Remove parameters that are not valid for the _search endpoint
+      def search_params
+        params = @params.merge(:_source => false)
+        params.select {|p, _| SEARCH_PARAMETERS.include? p}
+      end
+
+      # Remove parameters that are not valid for the _bulk endpoint
+      def bulk_params
+        @params.clone.tap { |p| p.delete(:q) }
```
look: We have ActiveSupport so you could use `Hash#except` here.

lib/elastomer/client/scroller.rb
```diff
@@ -120,7 +117,25 @@ def clear_scroll( scroll_ids )
...
+    # Returns the query as a hash
+    def add_sort_by_doc(query)
+      if query.nil?
+        query = {}
+      elsif query.is_a? String
+        query = MultiJson.load(query)
+      end
+
+      if query.has_key? :sort
+         raise ArgumentError, "#{query} cannot contain a sort #{query[:sort]}"
```
look: Since query could be quite large I think this exception is going to be hard to read. I'd suggest something like:

`Query cannot contain a sort (found sort '#{query[:sort]}' in query: #{query})`

With the query on the end.

lib/elastomer/client/delete_by_query.rb
juruen: Thank you for letting me know! I'll change it right away.

lib/elastomer/client/scroller.rb
juruen: Great suggestion, thanks.

lib/elastomer/client/delete_by_query.rb
look: Oops, I should have checked this more closely. Right now ActiveSupport is only a development dependency:

```
  spec.add_development_dependency "activesupport",      ">= 3.0"
```

I think you should leave it as-is. Sorry for not taking the time to check this out more closely.

lib/elastomer/client/delete_by_query.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_support/core_ext/hash'
```
juruen: @look I'm new to Ruby, so please, let me know whether this is how I'm supposed to use `ActiveSupport::Hash`. Thank you!

look:
I looked into the test failures more.

The `index_already_exists_exception` seems to be a timing issue -- there's a check for `exists?` and then an attempt to create the index. Perhaps the previous test deleted the index but the cluster state wasn't updated yet. If this continues to be a problem we could add a `wait_for_index_delete` method similar to `wait_for_index`.

The delete by query problem I ran into locally appears to be local state in the index that was persisted between tests (because the results have more than we would expect), but I could not reproduce it.

I've updated this branch with master and everything is passing so I'm going to merge it.

PR #164, opened 2017-12-01
Elireisman/es2 es5 version check

elireisman:
lifts version check from test harness into Elastomer::Client object for use in actual code shims

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -62,6 +62,24 @@ module Elastomer
       Semantic::Version.new(version)
     end
 
+    # Elasticsearch 2.0 changed some request formats in a non-backward-compatible
+    # way. Some tests need to know what version is running to structure requests
+    # as expected.
+    #
+    # Returns true if Elasticsearch version is 2.x.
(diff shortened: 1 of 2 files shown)


lib/elastomer/client.rb
```diff
@@ -62,6 +62,24 @@ def semantic_version
       Semantic::Version.new(version)
     end
 
+    # Elasticsearch 2.0 changed some request formats in a non-backward-compatible
+    # way. Some tests need to know what version is running to structure requests
+    # as expected.
+    #
+    # Returns true if Elasticsearch version is 2.x.
+    def on_es_version_2_x?
+      semantic_version >= "2.0.0" && semantic_version <  "3.0.0"
```
TwP: Three things here:

1. could we please remove the leading `on_` from this method
2. the regular old String `version` is fine - no need for `semantic_version`
3. double space after the less-than sign should be a single space

Same comments for the ES 5.X method below.

elireisman: opops sorry pushed too soon - will update

PR #165, opened 2017-12-01
 misc tweaks

elireisman:
shorten method name, whitespace fix, use cheaper version call/string comp internally for checks


elireisman:
This passes the tests locally against ES 2.4 and I played with the `version` vs `semantic_version` behavior in IRB and behaved reasonably for me. I'm gonna let this one simmer a bit to see if we want to make more adjustments before pushing unless I get a sign off. thanks!

PR #166, opened 2017-12-02
Add VersionSupport class for ES version differences

look:
VersionSupport is pretty basic right now but the idea is to have a place to
stash version-aware helpers. Currently it's used to provide a way to paper over
the difference between `text` and `keyword` fields in ES 5+ and `string` fields
in ES 2.

Supposedly ES 5 can take a legacy `string` field and do the right thing with it,
but there are severe limitations, such as not being able to handle an a `string`
field with `term_vectors` enabled. :(

See: https://github.com/elastic/elasticsearch/issues/21356

I've used the VersionSupport helpers to update all the tests that were using `string` fields. With this in place I'm getting 3 failures and 42 errors.

**UPDATE:** After some discussion with @TwP I updated the tests to use intention-revealing names for their version checks as well. I think this improves how the code reads nicely, but I wanted to ping you all so you can let me know what you think.

I also updated `VersionSupport` to raise an error if we attempt to use it with an unsupported version of Elasticsearch. This gives us a place to document what versions we support and also lets us blow up if we try to use it some place we shouldn't (like a 2.0 or 6.0 cluster).

cc: @elireisman @juruen

### DIFF
--- a/lib/elastomer/version_support.rb
+++ b/lib/elastomer/version_support.rb
@@ -1,0 +1,48 @@
+module Elastomer
+  class VersionSupport
+    attr_reader :client
+
+    def initialize(client)
+      @client = client
+    end
+
+    # COMPATIBILITY: Return a "text"-type mapping for a field.
+    #
+    # On ES 2.x, this will be a string field. On ES 5+, it will be a text field.
+    def text(**args)
+      reject_args!(args, :type, :index)
+
+      if client.es_version_2_x?
+        {type: "string"}.merge(args)
+      else
+        {type: "text"}.merge(args)
+      end
+    end
+
+    # COMPATIBILITY: Return a "keyword"-type mapping for a field.
+    #
+    # On ES 2.x, this will be a string field with not_analyzed=true. On ES 5+,
+    # it will be a keyword field.
+    def keyword(**args)
+      reject_args!(args, :type, :index)
+
+      if client.es_version_2_x?
+        {type: "string", index: "not_analyzed"}.merge(args)
+      else
+        {type: "keyword"}.merge(args)
+      end
+    end
+
+    private
+
+    # Internal: Helper to reject arguments that shouldn't be passed because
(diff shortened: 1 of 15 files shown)


lib/elastomer/version_support.rb
```diff
@@ -0,0 +1,48 @@
+module Elastomer
+  class VersionSupport
+    attr_reader :client
+
+    def initialize(client)
+      @client = client
+    end
```
TwP: My only thought at the moment is this ... could we pass in the version String here instead of a Client instance, and them move all version related checks into this class. This includes the `es_version_2_x?` and `es_version_5_x?` methods.

This change will make this class easier to test since it no longer depends upon a Client connected to a real Elasticsearch cluster. Passing in a String is much easier to stub out.

look: Yeah, good idea. Will update it.

elireisman:
I like this approach if y'all are happy with it I say 🚢  I can update my work to use these checks fairly easily

elireisman:
Actually, #167 PR is also gtg for ES5 but in order to surface all the percolate errors there, I had to mess with some `string` => `keyword/text` mappings in the tests. I might wait for this PR to merge so I can rebase mine on it rather than add my own branching, if this one is pushing soon?

test/version_support_test.rb
```diff
@@ -0,0 +1,84 @@
+require_relative './test_helper'
+
+describe Elastomer::VersionSupport do
+  describe "supported versions" do
+    it "allows 2.3.0 to 5.6" do
+      two_three_series = ["2.3.0", "2.3.5", "2.4.0", "2.9.9", "2.9.100"]
+      five_series = ["5.0.0", "5.0.9", "5.2.0", "5.6.9", "5.6.100"]
```
TwP: :heart_eyes_cat: very nice and exhaustive

look:
Thanks for the 🖊  off. Merging!

PR #167, opened 2017-12-02
Update Percolator APIs for ES 5.x upgrade

elireisman:
I expect this will receive a few more updated to the PR and probably a couple more test cases but thought I would put this out for everyone to have a peek in the meantime.

#### Changes for 5.x tl;dr:
- legacy `_percolate`, `_percolate/count` and `_mpercolate` API endpoints are deprecated but still function if percolation for the given index mapping is configured properly, and some perc queries are registered using the new API
- legacy `/:index/.percolate/:id` for registering perc q's on an index are gone in 5.x although docs claim ES5 can still read indices written by an older ES version
- However, types like `.percolator` starting with `.` aren't allowed anymore in 5.x
- an index that needs perc q's must register a new datatype for the queries in the index mapping, and that must have a field of type `percolator`
- new API registers/indexes perc queries by writing them to this new field in this new mapped data type

#### Further work
- we'll need to conditionally update the mappings at index create time in `github/github` and other uses of `elastomer-client` whenever they run on ES 5.x during the transition for indices that use percolator now
- We'll want to update the legacy endpoint usage (`_percolate` and friends) to use the new `_search` based perc API (using a `percolate` clause like [this](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-dsl-percolate-query.html#query-dsl-percolate-query)) once we're off ES2; this includes parsing changes for a search-style response

#### Questions:
*Parameterize the new perc query data type and it's field, as end users may name them differently in their own mappings?*
The new type and it's field can have any names but for this PR I chose what the docs do: `percolator` for the index data type, `query` for the field. If people get creative with those in their index mappings and we don't parameterize them in `elastomer-client`, end users will have a bad day. Let me know and I can add that plumbing here.

(more coming as I think of it)

### DIFF
--- a/lib/elastomer/client/percolator.rb
+++ b/lib/elastomer/client/percolator.rb
@@ -14,12 +14,7 @@ module Elastomer
         @id = client.assert_param_presence(id, "id")
 
         # COMPATIBILITY
-        @percolator_type =
-          if client.version_support.percolator_type?
-            'percolator'
-          else
-            '.percolator'
-          end
+        @percolator_type = client.version_support.percolator_type
       end
 
       attr_reader :client, :index_name, :id
--- a/lib/elastomer/client/percolator.rb
+++ b/lib/elastomer/client/percolator.rb
@@ -12,9 +12,18 @@ module Elastomer
         @client = client
         @index_name = client.assert_param_presence(index_name, "index name")
         @id = client.assert_param_presence(id, "id")
+
+        # COMPATIBILITY
+        @percolator_type = if client.es_version_5_x?
+                               'percolator'
+                           elsif client.es_version_2_x?
+                               '.percolator'
+                           else
+                               raise SupportedVersionError "elastomer-client doesn't support Percolate API for ES version < 2.x or > 5.x"
+                           end
       end
 
-      attr_reader :client, :index_name, :id
+      attr_reader :client, :index_name, :id, :percolator_type
 
       # Create a percolator query.
       #
@@ -25,7 +34,7 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def create(body, params = {})
-        response = client.put("/{index}/.percolator/{id}", defaults.merge(params.merge(:body => body, :action => "percolator.create")))
+        response = client.put("/{index}/{percolator_type}/{id}", defaults.merge(params.merge(:body => body, :action => "percolator.create")))
         response.body
       end
 
@@ -38,7 +47,7 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def get(params = {})
-        response = client.get("/{index}/.percolator/{id}", defaults.merge(params.merge(:action => "percolator.get")))
+        response = client.get("/{index}/{percolator_type}/{id}", defaults.merge(params.merge(:action => "percolator.get")))
         response.body
       end
 
@@ -51,7 +60,7 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def delete(params = {})
-        response = client.delete("/{index}/.percolator/{id}", defaults.merge(params.merge(:action => "percolator.delete")))
+        response = client.delete("/{index}/{percolator_type}/{id}", defaults.merge(params.merge(:action => "percolator.delete")))
         response.body
       end
 
@@ -69,7 +78,7 @@ module Elastomer
 
       # Internal: Returns a Hash containing default parameters.
       def defaults
-        {:index => index_name, :id => id}
+        {:index => index_name, :id => id, :percolator_type => percolator_type}
       end
 
     end  # Percolator
--- a/lib/elastomer/client/percolator.rb
+++ b/lib/elastomer/client/percolator.rb
@@ -19,7 +19,7 @@ module Elastomer
                            elsif client.es_version_2_x?
                                '.percolator'
                            else
-                               raise SupportedVersionError "elastomer-client doesn't support Percolate API for ES version < 2.x or > 5.x"
+                               raise IncompatibleVersionException "Percolator API not supported for ES #{@client.version}"
                            end
       end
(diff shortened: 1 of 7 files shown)


elireisman:
There's still some failing tests in `test/client/docs_test.rb` I uncovered after fixing some unrelated test errors, I'll update the patch Monday

test/test_helper.rb
```diff
@@ -106,6 +106,16 @@ def default_index_settings
...
 
+def default_index_mappings
+  if es_version_5_x?
+    {mappings: {percolator: {properties: {query: { type: "percolator"}}}}}
+  elsif es_version_2_x?
+    {}
+  else
+    raise IncompatibleVersionException "Percolator API not supported for ES #{$client.version}"
+  end
+end
```
TwP: I don't think we should define default index mappings for all tests in this fashion. The percolator definition only applies to the percolator tests; this code should be moved into the specific test case `setup` block.

test/client/multi_percolate_test.rb
```diff
@@ -26,6 +26,18 @@
...
+      # percolator clause must now be defined in index mappings as ordinary data type for ES versions >= 5.x
+      if es_version_5_x?
+          @index.update_mapping("percolator",
+            :percolator => {
+              :properties => {
+                :query => { :type => "percolator" }
+              }
+            }
+          )
+      end
```
TwP: This might create a race condition between the `create` call above and then the `udpate_mapping` call here. Another approach would be to create a `mappings` hash and then add the percolator section when testing under ES 5.X.

```ruby
mappings = {
   # doc1 and doc2 go here
}

if es_version_5_x?
  mappings[:percolator] = { ... }
end

@index.create \
  :settings => { ... },
  :mappings => mappings
```

Or something like that.

test/client/docs_test.rb
```diff
@@ -13,19 +13,24 @@
...
+              :title  => { :type => "text", :analyzer => "standard", :term_vector => "with_positions_offsets" },
+              :author => { :type => "keyword", :index => "not_analyzed" }
             }
           }
         }
 
+      # COMPATIBILITY
+      if es_version_5_x?
+        @index.update_mapping("percolator", { :properties => { :query => { :type => "percolator"}}})
+      end
```
TwP: Same comment here about creating a `mappings` hash and then calling `@index.create`.

elireisman:
This is now gtg for ES 5.6 but required mapping changes to surface all the percolator errors. I am planning to wait for #166 to :ship: so I can use that to shim some of my test mappings and get these tests to pass on ES 2.4 again, then should be ready for more eyes

test/client/docs_test.rb
```diff
@@ -463,6 +468,7 @@
     percolator2 = @index.percolator "2"
     response = percolator2.create :query => { :match => { :author => "defunkt" } }
     assert response["created"], "Couldn't create the percolator query"
+    @index.refresh
```
TwP: Why all the `@index.refresh` statements?

lib/elastomer/client/percolator.rb
```diff
@@ -12,9 +12,18 @@ def initialize(client, index_name, id)
...
         @id = client.assert_param_presence(id, "id")
+
+        # COMPATIBILITY
+        @percolator_type = if client.es_version_5_x?
+                               'percolator'
+                           elsif client.es_version_2_x?
+                               '.percolator'
+                           else
+                               raise IncompatibleVersionException "Percolator API not supported for ES #{@client.version}"
+                           end
```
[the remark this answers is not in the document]
elireisman: ah nice good to know, thanks! will update!

test/test_helper.rb
elireisman: ah good point will do

test/client/multi_percolate_test.rb
elireisman: excellent, will do!

test/client/docs_test.rb
elireisman: In my local env I found that the registered percolator queries weren't always visible when the section of the tests that submitted docs for percolation ran. Adding these made them work every time, but we could possibly do something more clever here to minimize refreshes-per-test? (`populate!` also calls a refresh etc.)

lib/elastomer/client/percolator.rb
```diff
@@ -25,7 +34,7 @@ def initialize(client, index_name, id)
       #
       # Returns the response body as a Hash
       def create(body, params = {})
-        response = client.put("/{index}/.percolator/{id}", defaults.merge(params.merge(:body => body, :action => "percolator.create")))
+        response = client.put("/{index}/{percolator_type}/{id}", defaults.merge(params.merge(:body => body, :action => "percolator.create")))
```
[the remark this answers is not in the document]
elireisman: ah excellent so that answers my question from above too - will update these calls

lib/elastomer/client/percolator.rb
```diff
@@ -12,6 +12,14 @@ def initialize(client, index_name, id)
         @client = client
         @index_name = client.assert_param_presence(index_name, "index name")
         @id = client.assert_param_presence(id, "id")
+
+        # COMPATIBILITY
+        @percolator_type =
+          if client.version_support.es_version_5_x?
```
look: Here's how I'd recommend doing this:

1. Add a new method to `VersionSupport` called `percolator_type` with this code.
2. Call it where ever you need the percolator type.

This way we can keep the conditional logic isolated.

I thought about making `es_version_2_x?` and `es_version_5_x?` private to discourage their use outside of `VersionSupport` but I thought we might need them for something.

elireisman: that makes sense, will update!

PR #168, opened 2017-12-04
pass analyzer as a json object

juruen:
In ES 5.x, it's required to pass the analyzer as
a json object in the body and not as a URI parameter.

This PR does so by passing all the parameters bar `:index` in the body.

\cc @look and @elireisman

### DIFF
--- a/lib/elastomer/client/index.rb
+++ b/lib/elastomer/client/index.rb
@@ -230,8 +230,7 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def analyze( text, params = {} )
-        body = {:text => text.to_s}.merge params
-        body.delete :index
+        body = text.is_a?(Hash) ? text : {text: text.to_s}
         response = client.get "{/index}/_analyze", update_params(params, :body => body, :action => "index.analyze")
         response.body
       end
--- a/lib/elastomer/client/index.rb
+++ b/lib/elastomer/client/index.rb
@@ -230,7 +230,8 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def analyze( text, params = {} )
-        response = client.get "{/index}/_analyze", update_params(params, :body => text.to_s, :action => "index.analyze")
+        body = params.merge(:text => text.to_s).tap { |h| h.delete(:index) }
+        response = client.get "{/index}/_analyze", update_params(params, :body => body, :action => "index.analyze")
         response.body
       end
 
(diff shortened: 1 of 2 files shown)


lib/elastomer/client/index.rb
```diff
@@ -230,7 +230,9 @@ def delete_alias( name, params = {} )
       #
       # Returns the response body as a Hash
       def analyze( text, params = {} )
-        response = client.get "{/index}/_analyze", update_params(params, :body => text.to_s, :action => "index.analyze")
+        body = {:text => text.to_s}.merge params
+        body.delete :index
+        response = client.get "{/index}/_analyze", update_params(params, :body => body, :action => "index.analyze")
```
TwP: Could we take a different approach here? Instead of merging `params` into the body, let's change `text` to accept either a String or a Hash. If a String is passed in then we adhere the current behavior. If a Hash is passed in, then we pass that as the body to the `/_analyze` endpoint. When a Hash is used, the user is responsible for populating it with the `text` field, the `analyzer`, etc.

I don't think this will break backwards compatibility since we still accept a simple String. Thoughts :thought_balloon:

juruen: @TwP I'm not sure I'm following. Is this what you are proposing?

```ruby
      def analyze( text, params = {} )
        if text.is_a? String
          body = {:text => text.to_s}.merge params
          body.delete :index
        else
           body = text
        end

        response = client.get "{/index}/_analyze", update_params(params, :body => body, :action => "index.analyze")
        response.body
      end
```

I'm not sure if by current behavior you mean master's or this PR's :)

juruen: Ideally, we would have the parameters that are sent as the request body in one variable, and other parameters in a separate variable. So we would achieve this while keeping the API backwards compatible.

juruen: The downside is to  end up with a slightly more confusing method.

look: I just fixed this myself in a different branch because I didn't realize this PR was doing the same thing. FWIW, I'd like to get this merged ASAP to avoid wasted duplication.

IMHO, changing input type of this method's parameters can be separated from fixing this bug, because that's enhancing the method's API which can be done later.

TwP: @juruen what I meant to communicate was this ...

```ruby
def analyze( text, params = {} )
  if text.is_a? Hash
    body = text
  else
    body = { text: text.to_s }
  end

  response = client.get "{/index}/_analyze", update_params(params, :body => body, :action => "index.analyze")
  response.body
end
```

juruen: @TwP IIRC, if we do that, when `text` is a string, the `analyze` parameters are passed in the URI, however you can't do that in  5.x, you need to pass them as a json object in the body.

PR #169, opened 2017-12-04
Remove deprecated `filtered` query type

look:
ES 2.0 deprecated the `filtered` query in favor of `bool` queries. ES 5.0
removed `filtered` queries entirely. This removes the test code using
`filtered`, except for the test of the query validator API.

The affected tests don't all pass because under ES 5 because delete by query uses the removed `scan` API which is fixed by #163. 

cc: @juruen @elireisman


elireisman:
This LGTM! I'm wondering since its a removal not a deprecation for ES5 if we should conditionally update the validator too, or if that's more in scope for the next parent issue and not the fix-the-tests round?

look:
I'm going to go ahead and merge this since @elireisman reviewed it.

PR #170, opened 2017-12-04
Change test using deprecated `filter` API

look:
`filter` was renamed to `post_filter` in 1.0(!) It has been removed in 5.0 so
this causes a test failure in the 5.x build.


elireisman:
LGTM :shipit: !

PR #171, opened 2017-12-05
Fixing transient failure

TwP:
The notification tests were failing when using Elasticsearch run via a Docker container. The test was running too quickly and the shards for the test index had not yet been allocated. Using our `wait_for_index` helper ensures the index has been fully created so the tests will pass.


elireisman:
This LGTM :shipit:

PR #172, opened 2017-12-05
Fix a couple test warnings

elireisman:


elireisman:
cool! thanks, will delete branch now

PR #173, opened 2017-12-05
Handle ES 5 removing `output` option for fields

look:
The `output` option for fields (to be used as the output for suggest queries)
was removed in 5.0. I split up the suggestion tests so there is one specifically
testing this behavior change.

See: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/breaking_50_suggester.html#_simpler_completion_indexing


elireisman:
LGTM!

PR #174, opened 2017-12-06
Add minitest-focus and pry-byebug to help with debugging

look:
With minitest-focus you can run a single test:

```
focus
it "is failing for some reason" do
  ...
end
```

With pry-byebug you can get an interactive debugger. I use it during test runs.

```
... code ...

binding.pry

... code ...
```

When `binding.pry` is reached, the interpreter will stop and drop you in a REPL.


elireisman:
Nice! Sorry was probably busy hand-editing in `byebug` to our dev deps myself to finish my ticket while you were submitting this :p but this arrangement looks much better. Definite +1 from me this stuff is super handy!

PR #175, opened 2017-12-06
Handle different query parsing exception between versions

look:
ES 2.x uses query_parsing_exception and ES 5.x uses query_shard_exception or
parsing_exception depending on where the parse error occurs (top-level keys
error out early, but a query string parse failure will happen on a shard and
throw a different exception).

Java classes:

https://github.com/elastic/elasticsearch/blob/v2.4.6/core/src/main/java/org/elasticsearch/index/query/QueryParsingException.java

https://github.com/elastic/elasticsearch/blob/v5.0.0-alpha1/core/src/main/java/org/elasticsearch/index/query/QueryShardException.java
https://github.com/elastic/elasticsearch/blob/v5.0.0-alpha1/core/src/main/java/org/elasticsearch/common/ParsingException.java

I also removed an ES 1.x code path that is unnecessary now.

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -306,21 +306,10 @@ module Elastomer
       raise ServerError, response if response.status >= 500
 
       if response.body.is_a?(Hash) && (error = response.body["error"])
-        # ES 2.X style
-        if error.is_a?(Hash)
-          root_cause = Array(error["root_cause"]).first || error
-          case root_cause["type"]
-          when "index_not_found_exception"; raise IndexNotFoundError, response
-          when "query_parsing_exception"; raise QueryParsingError, response
-          end
-
-        # ES 1.X style
-        elsif error.is_a?(String)
-          case error
-          when %r/IndexMissingException/; raise IndexNotFoundError, response
-          when %r/QueryParsingException/; raise QueryParsingError, response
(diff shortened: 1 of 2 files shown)


lib/elastomer/client.rb
```diff
@@ -306,21 +306,10 @@ def handle_errors( response )
...
-        elsif error.is_a?(String)
-          case error
-          when %r/IndexMissingException/; raise IndexNotFoundError, response
-          when %r/QueryParsingException/; raise QueryParsingError, response
-          when %r/ParseException/; raise QueryParsingError, response
-          end
+        root_cause = Array(error["root_cause"]).first || error
+        case root_cause["type"]
+        when "index_not_found_exception"; raise IndexNotFoundError, response
+        when *version_support.query_parse_exception; raise QueryParsingError, response
```
elireisman: concise use of the splat here, I'm taking Ruby notes...!

elireisman:
🚢 lgtm!

lib/elastomer/client.rb
look: Hah, I had to look up how to do multiple matches in a case statement, and it turned out splat is the way! https://bengribaudo.com/blog/2013/12/31/2530/ruby-case-testing-against-arrays-of-values

PR #176, opened 2017-12-06
Update multi_percolate tests

elireisman:
A subset of multi_percolate tests in `index_tests.rb` that exercises the block form of bulk percolation still had an error. Exploration with `byebug` and `awesome_print` revealed that params weren't populating the final request body in the same way they did with the full-body tests or my tests of the legacy `_mpercolate` APIs that worked on 2.4 and 5.6 locally.

This PR updates all that stuff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
-# WIP for 3.3.0 release
+# WIP for 3.0.0 release
 - Fixed swapped args in {Client,Index}#multi\_percolate count calls using block API
-
 
 ## 2.3.0 (2017-11-29)
 - Remove Elasticsearch 1.x and earlier code paths
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# WIP for 3.3.0 release
+- Fixed swapped args in {Client,Index}#multi\_percolate count calls using block API
+
+
 ## 2.3.0 (2017-11-29)
 - Remove Elasticsearch 1.x and earlier code paths
 - Fix CI and configure an Elasticsearch 5.6 build
--- a/lib/elastomer/client/multi_percolate.rb
+++ b/lib/elastomer/client/multi_percolate.rb
@@ -80,8 +80,7 @@ module Elastomer
       #
       # Returns this MultiPercolate instance.
       def percolate(doc, header = {})
-        merged = @params.merge(header)
-        add_to_actions(:percolate => merged)
+        add_to_actions(:percolate => @params.merge(header))
         add_to_actions(:doc => doc)
       end
 
@@ -94,8 +93,7 @@ module Elastomer
       #
       # Returns this MultiPercolate instance.
       def count(doc, header = {})
-        merged = @params.merge(header)
-        add_to_actions(:count => merged)
+        add_to_actions(:count => @params.merge(header))
         add_to_actions(:doc => doc)
       end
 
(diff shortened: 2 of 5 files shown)


test/client/index_test.rb
```diff
@@ -367,11 +372,12 @@
 
       @index.percolator("1").create :query => { :match_all => { } }
       @index.percolator("2").create :query => { :match => { :author => "pea53" } }
+      @index.refresh
 
       h = @index.multi_percolate(:type => "doco") do |m|
         m.percolate :author => "pea53"
         m.percolate :author => "grantr"
-        m.count({}, { :author => "grantr" })
+        m.count({ :author => "grantr" }, {})
```
look: What's the deal with flipping these hashes in the params list? Does it represent an API change?

lib/elastomer/client/multi_percolate.rb
```diff
@@ -80,7 +80,8 @@ def initialize(client, params = {})
       #
       # Returns this MultiPercolate instance.
       def percolate(doc, header = {})
-        add_to_actions(:percolate => header)
+        merged = @params.merge(header)
```
look: FWIW, I'd probably call this like `add_to_actions(percolate: @params.merge(header))` since `#merge` returns a new Hash. It's personal preference, though.

elireisman: ah yeah I just did that for clarity of what I was changing but more concise is good I agree, will update!

test/client/index_test.rb
elireisman: Yeah this was a wierd one!

After adding the param merges I noticed in `byebug` that this `count` call was generating a bad output (injecting the `doc` into the params/`percolate` bulk line and a blank doc into the for the `count`)

The API docs in the [code](https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/index.rb#L496) indicate these were backwards and should match the format of the `percolate` (in which the 2nd arg params hash is optional).

I see no uses of `multi_percolate` in `github/github` so I'm guessing this just silently passed it's test and went unused all this time? Line Git-blames to 2015, so a bit scary, but the output looks right now, and is [wrong on master](https://gist.github.com/2909d8aa40479a7cf0b424bb9304a0b2) so little creepy to flip it I agree, but seems legit? Happy to investigate further if anyone suspects I got turned around here and am being too brave :)

look: Whoa crazy! Unfixed bug from way back. Since we're going to bump the major version anyway, I think a breaking change is OK. Can you add it to the CHANGELOG as part of this PR so we don't forget about it?

test/client/index_test.rb
```diff
@@ -431,8 +437,8 @@
           @index.docs.index(document)
         end
 
-        assert_equal 400, exception.status
-        assert_match /\[output\]/, exception.message
+        assert_equal( 400, exception.status)
```
look: Yeah this kind of thing is why I prefer parenthesis more often than not. Great commit message 😂 

Sorry to style snipe you but do you mind deleting the space after the parens here?

elireisman: ah crap yeah you got it thanks!

test/client/index_test.rb
elireisman: Good call, will update now!

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# WIP for 3.3.0 release
```
look: Can you make this 3.0.0 before merging? Thanks!

elireisman: Gah yes thanks!

PR #177, opened 2017-12-06
Dockerized Janky builds

TwP:
The goal of this PR is to use Docker in the Janky build environment so that we can swap out Elasticsearch versions with ease.


TwP:
Could I get some help from @github/dx on this Docker configuration out on the Janky hosts.

The `docker-compose.es24.yml` file is configured to spin up an Elasticsearch 2.4.6 cluster inside a Docker container, and it will forward port 9200 from the container back out to port 19200 on the host machine. The problem is that any connections on port 19200 on the host machine are being reset by peer:

```sh
> curl -XGET '127.0.0.1:19200' 
curl: (56) Recv failure: Connection reset by peer
```

But if I attach a bash process to the running Docker container, I can easily `curl` the Elasticsearch process running on port 9200.

There is a `docker-proxy` process running on the host that looks like this in a `ps` output:

```
/usr/bin/docker-proxy -proto tcp -host-ip 127.0.0.1 -host-port 19200 -container-ip 172.18.0.2 -container-port 9200
```

So I guess the port should be forwarding correctly to the container, but something is amiss. What am I missing here on this janky host, and what do I need to do to get this port forwarding working properly? Any help or insight would be greatly appreciated.

script/poll-for-es
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#
+# This script will poll the Elasticsearch health endpoint until the cluster
+# reaches a yellow state which is good enough for testing. This script will poll
+# for up to 30 seconds waiting for Elasticserach to start. It will give up at
```
look: s/Elasticserach/Elasticsearch/

script/poll-for-es
```diff
@@ -0,0 +1,17 @@
...
+# that time and return a non-zero exit code.
+
+es_port=${ES_PORT:-9200}
+count=0
+
+until $(curl -s "localhost:${es_port}/_cluster/health?wait_for_status=yellow&timeout=30s" > /dev/null 2>&1); do
```
look: This is handy. I might switch to using this on Travis CI. I used `wget` with its built-in retries but it's a little opaque how long it will try for.

TwP:
> So is the plan to get this working with ES 2.4 and then create a new compose file for ES 5.6?

I have a compose file for ES 5.6 ready to go on my computer here. This PR is my attempt to spike out what a test build on ES 5.6 could look like for our main application. If this goes well, then I can follow the same approach over there.

TwP:
Hey @github/dx - my problem was that I neglected to include the default network setup for the Docker containers. After that was included in the `cibuild` docker compose file, everything worked swimmingly.

@look / @elireisman this code is ready to be merged

elireisman:
This is great, sorry I didn't dig into it more sooner!

PR #180, opened 2017-12-06
Fix the `delete_by_query` logic

TwP:
The change here allows `index` and `type` to be passed to the scroll query. These parameters are used to restrict the `delete_by_query` action to the given index and document type.

### DIFF
--- a/lib/elastomer/client/delete_by_query.rb
+++ b/lib/elastomer/client/delete_by_query.rb
@@ -48,28 +48,31 @@ module Elastomer
       DeleteByQuery.new(self, query, params).execute
     end
 
-    SEARCH_PARAMETERS = [
-      :q,
-      :df,
-      :analyzer,
-      :analyze_wildcard,
-      :batched_reduce_size,
-      :default_operator,
-      :lenient,
-      :explain,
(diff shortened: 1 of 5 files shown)


lib/elastomer/client/delete_by_query.rb
```diff
@@ -48,29 +48,32 @@ def delete_by_query(query, params = {})
...
-      :terminate_after,
-      :from,
-      :size,
-      :search_type,
-      :scroll
-    ].to_set.freeze
-
     class DeleteByQuery
 
+      SEARCH_PARAMETERS = %i[
```
look: I just learned about `%i` today. So nice!

lib/elastomer/client/delete_by_query.rb
```diff
@@ -48,29 +48,32 @@ def delete_by_query(query, params = {})
...
+        index
+        type
```
look: To be clear, these were the two parameters that were missing, right?

TwP: Correct. These were being stripped out of the params passed to the scroll query.

PR #181, opened 2017-12-06
CI build environment tweaks

TwP:
These are cosmetic changes to the output of the cibuild script so that it looks nicer in the janky environment. Also disabling more X-Pack features in the ES 5.6 Docker image.


elireisman:
LGTM 🐑 it!

PR #182, opened 2017-12-07
Fail fast on known invalid indexing directive fields

### LINKED ISSUE #178: Handle version differences with URL parameter fields
Certain document hash keys are "special" and get converted into URL parameters rather than being indexed with the document.

These fields are added to a document. They get pulled out and added to the query
string (rather than being passed in a a parameter to the index method). Any unknown underscore fields are ignored and indexed as usual.

https://github.com/github/elastomer-client/blob/9e96707e74ab890607f5cd795cf186a9af457898/lib/elastomer/client/docs.rb#L520

These URL parameters are different between versions. `_consistency` got re-implemented as `_wait_for_active_shards` and has different values, so it's not a direct mapping. 

Additionally, ES 5 no longer allows unknown URL parameters. So we need to not send it or force the caller to deal with the error. And vice-versa, there is a problem as well: if someone uses a new
parameter from ES 5, it will end up in a document as a field unless we pull it out and put it on the URL as a parameter (which ES 2 will ignore), or pull it out and ignore it.

We need to decide how to handle these differences in the library.

References:

* [ES 5 `wait_for_active_shards`](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/docs-index_.html#index-wait-for-active-shards)
* [ES 2 `consistency`](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/docs-index_.html#index-consistency)
* [Issue about the change](https://github.com/elastic/elasticsearch/issues/21804)

look:
Pull the different "special" fields that affect indexing into VersionSupport in
order to return a different Hash of parameter name/field name depending on the
client's version.

If a caller attempts to index a document with a known indexing directive field
that is not supported by the version of Elasticsearch, the client will fail fast
and not allow the document to be indexed.

Fixes #178.

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -517,9 +517,6 @@ Percolate
         client.multi_percolate(params, &block)
       end
 
-      SPECIAL_KEYS = %w[index type id version version_type op_type routing parent timestamp ttl consistency replication refresh].freeze
-      SPECIAL_KEYS_HASH = SPECIAL_KEYS.inject({}) { |h, k| h[k.to_sym] = "_#{k}"; h }.freeze
-
       # Internal: Given a `document` generate an options hash that will
       # override parameters based on the content of the document. The document
       # will be returned as the value of the :body key.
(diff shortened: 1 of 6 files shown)


look:
With this change, tests are now passing against ES 5.6. 🎉

elireisman:
No complaints here, looks great :shipit:

lib/elastomer/client/docs.rb
```diff
@@ -534,10 +531,23 @@ def from_document( document )
...
+          # COMPATIBILITY
+          # Fail fast if a consumer is attempting to use an indexing parameter
+          # for a different version of Elasticsearch. Elasticsearch 5+ is strict
+          # about parameter names so we need to either ignore these (in which
+          # case they would be indexed with the document) or fail-fast.
+          # Elasticsearch 2.X will happily ignore unknown parameters, but we
+          # felt it was best to consistently fail fast.
+          client.version_support.unsupported_indexing_directives.each do |key, field|
+            if document.key?(field) || document.key?(field.to_sym)
+              raise IncompatibleVersionException, "Elasticsearch #{client.version} does not support the '#{key}' indexing parameter"
```
TwP: My thought here is that the `IncompatibleVersionException` should be used for features that are not available across both major versions of Elasticsearch that we support. For parameters, we should create a new `IllegalArgumentException` that is raised when an incompatible parameter is used. This mirrors the exception returned by Elasticsearch itself, and we will be using this exception across more and more the client library.

The distinction helps us know when a feature is just gone versus when a request needs a little tweaking.

look: Yeah, I agree. I'll create a new exception for this.

elireisman: this makes sense :+1:

elireisman:
+1 for the new exception, will come in handy for the rest of this effort as @TwP mentioned
➡️ 🚢 ⬅️

look:
woohoo!

PR #184, opened 2017-12-07
Raise an IllegalArgument exception

TwP:
With Elasticsearch 5.X now validating the parameters passed with each HTTP request, this code raises an InvalidParameter excpetion if we violate what Elasticsearch expects. This will help us find and fix these problems in our own application.

### DIFF
--- a/lib/elastomer/client/errors.rb
+++ b/lib/elastomer/client/errors.rb
@@ -109,7 +109,7 @@ module Elastomer
     IncompatibleVersionException = Class.new Error
 
     # Exception for client-detected invalid Elasticsearch parameter
-    InvalidParameter = Class.new Error
+    IllegalArgument = Class.new Error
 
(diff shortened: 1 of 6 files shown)


look:
Ah, I didn't realize that Elasticsearch's exception was literally called `illegal_argument_exception`. If you want to harmonize the exception name like we do for the others, go for it!

lib/elastomer/client/errors.rb
```diff
@@ -109,7 +109,7 @@ def fatal
     IncompatibleVersionException = Class.new Error
 
     # Exception for client-detected invalid Elasticsearch parameter
```
look: Do you mind updating this comment since it's now client-detected AND server response?

TwP: not at all

PR #185, opened 2017-12-08
Implement native _delete_by_query for ES 5+

look:
This is a first pass at implementing native `_delete_by_query` support for the library.

Because the ES 5 API is pretty different than what we implemented in the client itself, I added a new method for the new way of doing things.

This is a little rough still. I wanted to get it up so we could discuss if we even what to do it this way. If so I'll clean up the docs/tests/validations more.


look:
cc: @elireisman too.

One thing I'm not sure if there will be a way to test is failures. The response includes a list of failed deletes, but I'm not sure how to cause one. The previous code faked out a `_bulk` API request with some failures but we won't be able to do that with this version.

elireisman:
@look this LGTM! I am finishing up the `/_tasks` API work and same thing, there are a couple things that will be tricky to test without mocks or a background thread causing lots of fake work.

In my local experiments the tasks are ephemeral and too quickly completed to call back against a single task ID before it's off the list/unregistered and treated like it never existed. In curl experiments, even fairly heavy indexing of individual docs in the background isn't catching the individual indexing tasks while they are still on the list. I'll push the PR soon after I relocate again and maybe we can talk test strategies for all this stuff?

look:
I think any operation that's happening entirely inside Elasticsearch (that is, not as a series of HTTP calls between the client and the server) we are not really going to be able to mock. 

I could mock an HTTP response with failures in it, but it wouldn't prove much because all the method does is return the parsed JSON from the response.  ¯\_(ツ)_/¯

TwP:
I like having a new native method :+1:. Following in that vein, it would be nice to have an `app_delete_by_query` that provides the current implementation we already have. And then the current `delete_by_query` method in the client can choose the native method if it is supported by the ES version and then fall back to the app implementation.

I'm not concerned about the difference in return value structure. We will need to account for that in the application anyways. We can implement that when we need to inspect the results.

Those are all my thoughts on this one.

look:
Thanks for the suggestion, I have done this and I think it looks pretty good. I added a lot of docs to warn users about the behavior switching based on ES version and deprecated the `app_delete_by_query` method.

look:
Bombs away 💥

PR #186, opened 2017-12-08
Recycle socket connection on error

TwP:
This change recycles the Faraday connection if an `OpaqueId` error is encountered. The one problem with this solution is that any persistent connections will be left open because Faraday does not provide a nice way to call `close` on the underlying socket.


elireisman:
Read the enterprise issue and took a look at this. My instinct is to worry about those persistent connections piling up as @look mentioned, but I'm still building up an idea of what Faraday might expose to deal with this, how we're tracking these opaque ID's internally, _and_ I have no intuition yet for how disruptive it might be to customers to include a fixed timeout in the ES queries for GHE deployments (my instinct is this won't be a great experience on their end)  so I'm not sure I'm going to have anything useful to add without some more research and thinking at this point :(

look:
This may be the best we can do, so I don't have an objection to merging it.

However considering the current state of Faraday and our own usage of this library I think we should consider if it is buying us anything to use Faraday. We could use Excon directly and have more control over this kind of thing.

elireisman:
hey just in case you're waiting on me to follow up on this - I'm fine with merging it too if you fellows are!

Issue #158, opened 2017-11-30
Incorporate new ES 5.X features


TwP:
We need to pull in new features from ES 5.X that are not yet supported by the gem. When the gem is used with an ES 2.X cluster, a sensible empty result should be returned from these methods.

- [x] Delete by query [docs](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/docs-delete-by-query.html) @look #185
- [x] Task management API [docs](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/tasks.html) @elireisman #187 
- [x] Raise `IllegalArgument` exceptions when returned from Elasticsearch #184 
- [x] Upgrades tp + github/github deployment(s) of `elastomer-client` to pull in `VersionSupport` shims for ES2~ES5 transition period @elireisman @look (multiple patches)
- [x] Pattern for parameter validation @twp [here](https://github.com/github/elastomer-client/pull/198)

elireisman:
With the release of `elastomer-client` v3.1.x I think we're all done here, including param validation - closing this

PR #187, opened 2017-12-08
Add Tasks API support (for ES 5.x or newer) to elastomer-client

elireisman:
This adds basic API support for the Tasks endpoints for ES 5.x or newer, and some (very!) basic tests. This follows the pattern of creating a client-bound convenience object from the `Elastomer::Client` (such as an `Index`) for calling the API-specific family of methods. In this case `client.tasks` gets you a `Tasks` upon which you can call things like `get` or `cancel_by_id`, and so on. 

#### Outstanding Issues:
1. The elastic.co docs [here](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/tasks.html) document the `*_by_id` calls wrong!

^ Instead of supplying a numeric task ID as part of the path segment `task_id:NUMBER`, the proper API calls include a path segment `<node_id>:<task_id>` where both are identifiers for a single ES cluster node and an integer task ID for the given task. These can be viewed when curling the generic `/_tasks` endpoint and looking under any `nodes` entry for the _keys_ to its child `tasks` object.

2. Turns out in local testing, tasks are _very_ ephemeral and fast-running so I don't test much of the happy path (yet)

^ Obtaining a task ID that's still valid long enough to make calls against turns out to be really tricky (or I'm really dumb) and so I think the options are: don't cover the happy path well, use mocking here, or generate fake work in a background thread of the test suite sufficient to get your mitts on a valid task ID you can test against. So far I've had bad luck with that locally. Ideas welcome!


elireisman:
Hmm kinda stumped why [this](https://janky.githubapp.com/8860210/output#output) janky build on es56 is failing - works great on my local ES 5.6.4, this same test worked before (!) and its a ruby level error, where no `sum` method on `Array`, which I'm calling like so on the results of a `map` call:
`total_tasks = h["nodes"].map { |k, v| v["tasks"].keys.count }.sum`

^ which seems to work as long as the results of the map call is an Array of integers, or empty in my local env. I'm on Ruby 2.4.1 locally, didn't see in the dump if maybe CI is on a different version or something?

Any idea what I did here?

lib/elastomer/client.rb
```diff
@@ -69,6 +69,15 @@ def info
...
 
+    # Returns a Tasks instance for querying the cluster bound to this client for
+    # metadata about internal tasks in flight, and to submit administrative
+    # commands concerning those tasks.
+    #
+    # Returns a new Tasks object associated with this client
+    def tasks
+      Tasks.new(self)
+    end
+
```
TwP: Could you move this method into the `lib/elastomer/client/tasks.rb` file, please. The convention here is that an API endpoint definition includes all the methods for that endpoint. We do something similar in the `lib/elastomer/client/index.rb` file.

lib/elastomer/client/tasks.rb
```diff
@@ -0,0 +1,169 @@
...
+      #
+      # Examples
+      #
+      #   tasks.cancel_by_id "DmteLdw1QmSgW3GZmjmoKA", 123
+      #   tasks.cancel_by_id "DmteLdw1QmSgW3GZmjmoKA", 456, :pretty => true
+      #
+      # Returns the response body as a Hash
+      def cancel_by_id(node_id, task_id, params = {})
+        raise IllegalArgument, "no node ID provided" if node_id.to_s.empty?
+        raise IllegalArgument, "no task ID provided" unless task_id.to_i
```
TwP: The `to_i` method in Ruby works for **everything**.

```ruby
nil.to_i  #=> 0
```

If you want to ensure that the user is passing in a Numeric value, then you can check the type of the `task_id`.

```ruby
raise ArgumentError, "invalid task ID provided: #{task_id.inspect}" unless task_id.is_a?(Integer)
```

The `ArgumentError` here is a Ruby defined exception signaling that the wrong arguments were passed to the method. It is more suitable in this case.

elireisman:
Update:

Not having much luck faking a slow task, trying variants of [this](https://gist.github.com/6eb2fea53ca0a7ea33f3ec1875263547) with curl and in the test case as shown here. Returns right away before my list check fires, and even if the script was running async on the ES side, nothing related shows up in the task list when checked. 

This somewhat mirrors other odd behavior I was seeing last night while doing heavy indexing locally where almost all of my unfiltered `/_tasks` queries showed no indexing tasks in flight during hours of indexing dummy docs...I see Painless queries that run too long also get reaped on the ES side as a circuit breaker, so it may be even if I get this behaving reasonably, we'll have to maintain a delicate balance between running long enough to detect the task and short enough not to get reaped which might get env CI.

I'll keep banging on this and playing with scripted queries + other types of fake work generation, but might have to explore other opts if this doesn't pan out. Onward! 😣

test/client/tasks_test.rb
```diff
@@ -0,0 +1,93 @@
+require File.expand_path("../../test_helper", __FILE__)
```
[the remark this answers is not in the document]
elireisman: Sounds good. I stole this from similar test cases that use it - can switch them all to `require_relative` if you like while I'm in there?

test/client/tasks_test.rb
```diff
@@ -0,0 +1,93 @@
...
+  before do
+    unless $client.version_support.es_version_5_x?
+      skip "Tasks API is not supported in ES version #{$client.version}"
+    end
+
+    @tasks = $client.tasks
+  end
+
+  after do
+    pass # nothing to do here yet...
```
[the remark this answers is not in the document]
elireisman: Ah ok wasn't sure if empty block was OK here or if I should just have no `after` block at all? Will end up using this if I can get Painless scripting to work

lib/elastomer/client/tasks.rb
elireisman: Thank makes sense, thanks, will switch to `Numeric` type check. Was hoping to do something that allows for `"123"` and `123` but not `"joe"` will think on it.

lib/elastomer/client.rb
elireisman: Ah I see thanks will do!

lib/elastomer/client/tasks.rb
look: You can use `Integer()` for that:

```
irb(main):001:0> Integer("1")
=> 1
irb(main):002:0> Integer("1.0")
ArgumentError: invalid value for Integer(): "1.0"
	from (irb):2:in `Integer'
	from (irb):2
	from /Users/look/.rbenv/versions/2.4.2/bin/irb:11:in `<main>'
irb(main):003:0> Integer(1)
=> 1
irb(main):004:0> Integer("joe")
ArgumentError: invalid value for Integer(): "joe"
	from (irb):4:in `Integer'
	from (irb):4
	from /Users/look/.rbenv/versions/2.4.2/bin/irb:11:in `<main>'
```

elireisman:
This should be basically gtg at this point: post review stuff is updated, and after much experimenting I have a test that generates enough work in a background thread to be able to list then verify-by-ID a task reliably, and without so much background work that it slows down the suite. So we are now excercising `get_by_id` and friends properly.

As of now, I was not able to trick ES5 into giving me a task categorized as `cancellable` that I could similarly exercise `cancel_by_id` on in a reasonable amount of running time.

test/client/tasks_test.rb
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+
+  it "successfully waits for task to complete when wait_for_completion and timeout flags are set" do
+    # do some busy work in bkg thread to keep some tasks running longer
+    Thread.new do
+      begin
+        name = "elastomer-tasks-tests".freeze
+        index = $client.index(name)
+        wait_for_index(name)
```
TwP: Two things here for this test case:

1.  The thread here should create its own connection to the Elasticsearch cluster and use that in the test. The code outside the thread will also be sending requests which might stomp on the requests coming from inside the thread.

2. You need to `create()` the index first before calling the `wait_for_index` method.

```ruby
client = Elastomer::Client.new \
  host:         $client.host,
  port:         $client.port,
  read_timeout: $client.read_timeout,
  open_timeout: $client.open_timeout

index = client.index(name)
index.create({})
wait_for_index(name)
```

test/client/tasks_test.rb
```diff
@@ -0,0 +1,179 @@
...
+      end
+    end
+
+    # locate task(s) long running enough to be looked up by ID
+    target_tasks = []
+    3.times.each do
+      target_tasks = @tasks.get["nodes"]
+        .map { |k, v| v["tasks"] }
+        .flatten.map { |ts| ts.select { |k, v| /health/ =~ v["action"] } }
+        .flatten.reject { |t| t.empty? }
```
TwP: I would extract all of this logic into a helper method that you put at the end of this test file. Something like ...

```ruby
def lookup_health_tasks(timeout: 5)
  tasks = $client.tasks
  start = Time.now

  ary = []

  loop do
    break if (Time.now - start) >= timeout

    ary = tasks.get["nodes"]
      .map { |k,v| v["tasks"] }.flatten
      .map { |ts| ts.select { |k,v| /health/ =~ v["action"] }}.flatten
      .reject { |t| t.empty? }

    break unless ary.empty?
    sleep 0.250
  end

  ary
end
```

And then you can use this method in your test to lookup the `target_tasks` array.

test/client/tasks_test.rb
```diff
@@ -0,0 +1,179 @@
...
+    assert success
+  end
+
+  it "locates the task properly by ID when valid node and task IDs are supplied" do
+    # do some busy work in bkg thread to keep some tasks running longer
+    Thread.new do
+      begin
+        name = "elastomer-tasks-tests".freeze
+        index = $client.index(name)
+        wait_for_index(name)
```
TwP: Same comment here regarding a new `cilent` for this thread to use. Hrmmm ... a helper method to duplicate a client connection would be super rad. Better yet, implementing the `Elastomer::Client#dup` method would be super helpful. I can do that work in a separate PR.

For now using a different `client` is recommended.

test/client/tasks_test.rb
```diff
@@ -0,0 +1,179 @@
...
+    # locate task(s) long running enough to be looked up by ID
+    target_tasks = []
+    3.times.each do
+      target_tasks = @tasks.get["nodes"]
+        .map { |k, v| v["tasks"] }
+        .flatten.map { |ts| ts.select { |k, v| /health/ =~ v["action"] } }
+        .flatten.reject { |t| t.empty? }
+      break if target_tasks.size > 0
+    end
+    assert !target_tasks.empty?
```
TwP: You can use the `lookup_health_tasks` helper method here, too.

elireisman:
OK rebased to master, merged a bunch of changes, and tested locally on 2.4 and 5.6 I think we're finally in good shape here, thanks again for the careful reviews! 🍻

elireisman:
One piece of unfinished business here is param validation. I have a whitelist in the Tasks class, but saw in #158 that there's a task for normalizing param validation so I figured I'd wait and merge that in before merging this. Alternately, I can just add it to this PR for Tasks

lib/elastomer/version_support.rb
```diff
@@ -37,6 +37,12 @@ def supports_warmers?
       es_version_2_x?
     end
 
+    # COMPATIBILITY: The Tasks API is evolving quickly; features, and request/response\
```
look: looks like there's a extra `\` in this comment

elireisman: thanks!

PR #188, opened 2017-12-11
Implement `dup` method for Client

TwP:
The goal of this PR is to implement an `Elastomer::Client#dup` method that returns a client with the same configuration but with a unique connection to the Elasticsearch cluster. This is useful for multi-threaded environments where the socket connection to the cluster should not be shared across thread boundaries.


elireisman:
Thanks for this!

PR #189, opened 2017-12-12
Don't use test helper methods in production code

look:
Oops, I used a method defined in test_helper.rb in this class. So of course all
the tests passed, but when I tried to use it in real code... 💣💥


elireisman:
LGTM 🚢 !!!

PR #190, opened 2017-12-14
Retry requests when possible

TwP:
The goal of this PR is to add automated retries for failed requests. A new attribute `max_retries` is added to the client. This determines the maximum number of retry attempts that will be made by the client; it can be overridden on a per-request basis.

Only `GET` and `HEAD` requests can be retried. My thought here is that `PUT`, `POST`, and `DELETE` requests might actually succeed on the Elasticsearch cluster, but the request itself might timeout. In this case the desired modification was made so we do not want to retry the request.

Only a handful of exceptions can be retried. See:

https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/errors.rb#L91-L95

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -30,9 +30,10 @@ module Elastomer
     #   :opaque_id    - set to `true` to use the 'X-Opaque-Id' request header
     #   :max_request_size - the maximum allowed request size in bytes (defaults to 250 MB)
     #   :max_retries      - the maximum number of request retires (defaults to 0)
+    #   :retry_delay      - delay in seconds between retries (defaults to 0.075)
     #
     def initialize(host: "localhost", port: 9200, url: nil,
-                   max_retries: 0, read_timeout: 5, open_timeout: 2,
+                   read_timeout: 5, open_timeout: 2, max_retries: 0, retry_delay: 0.075,
                    opaque_id: false, adapter: Faraday.default_adapter, max_request_size: MAX_REQUEST_SIZE)
 
       @url = url || "http://#{host}:#{port}"
@@ -41,16 +42,18 @@ module Elastomer
       @host = uri.host
       @port = uri.port
 
-      @max_retries      = max_retries
       @read_timeout     = read_timeout
       @open_timeout     = open_timeout
+      @max_retries      = max_retries
+      @retry_delay      = retry_delay
       @adapter          = adapter
       @opaque_id        = opaque_id
       @max_request_size = max_request_size
     end
 
     attr_reader :host, :port, :url
-    attr_reader :max_retries, :read_timeout, :open_timeout, :max_request_size
+    attr_reader :read_timeout, :open_timeout
+    attr_reader :max_retries, :retry_delay, :max_request_size
 
     # Returns a duplicate of this Client connection configured in the exact same
     # fashion.
@@ -232,7 +235,7 @@ module Elastomer
           error = wrap_faraday_error(boom, method, path)
           if error.retry? && RETRYABLE_METHODS.include?(method) && (retries += 1) <= request_max_retries
             params[:retries] = retries
-            sleep 0.075
+            sleep retry_delay
             retry
           end
           raise error
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -15,8 +15,8 @@ module Elastomer
 
     # Create a new client that can be used to make HTTP requests to the
     # Elasticsearch server. If you use `:max_retries`, then any GET or HEAD
-    # request will be retried up this many times with a 75ms delay between the
-    # retry attempts. Only non-fatal exceptions will retried automatically.
+    # request will be retried up to this many times with a 75ms delay between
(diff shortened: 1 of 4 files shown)


lib/elastomer/client.rb
```diff
@@ -23,25 +25,28 @@ class Client
     #   :adapter      - the Faraday adapter to use (defaults to :excon)
     #   :opaque_id    - set to `true` to use the 'X-Opaque-Id' request header
     #   :max_request_size - the maximum allowed request size in bytes (defaults to 250 MB)
+    #   :max_retries      - the maximum number of request retires (defaults to 0)
     #
-    def initialize( opts = {} )
-      host = opts.fetch :host, "localhost"
-      port = opts.fetch :port, 9200
-      @url = opts.fetch :url,  "http://#{host}:#{port}"
+    def initialize(host: "localhost", port: 9200, url: nil,
```
miguelff: Much better with an explicit signature! 💯

lib/elastomer/client.rb
```diff
@@ -228,6 +238,20 @@ def request( method, path, params )
...
+    # Faraday error. A generic Error is returned if we cannot wrap the given
+    # Faraday error.
+    #
+    #   error  - The Faraday error
+    #   method - The request method
+    #   path   - The request path
+    #
+    def wrap_faraday_error(error, method, path)
+      error_name  = error.class.name.split("::").last
+      error_class = Elastomer::Client.const_get(error_name) rescue Elastomer::Client::Error
```
miguelff: Does this get a specific class matching the server's Java exception?

elireisman: I think it's mapping back to one of [these](https://github.com/github/elastomer-client/blob/master/lib/elastomer/client/errors.rb#L76-L84)

elireisman:
:+1: and good learnings for me wrt Faraday/Excon !

look:
@TwP what do you think about adding some backoff for the retries?

TwP:
> what do you think about adding some backoff for the retries?

I'd like the end user to handle that level of finesse. The goal here is just a quick and simple retry mechanism that has predictable timing. Backoff is wonderful, but the end user will be blocked until that backoff period expires.

This is aimed more at re-establishing a connection if the host disappears or is disconnected for whatever reason.

look:
It will try to re-establish the connection very quickly though. If whatever happened hasn't fixed itself pretty much instantly this will fail. Whenever I've implemented retries I've always needed to add a bit of backoff. You could give the user control by providing a `max_duration` parameter and stopping the retries after that amount of time has elapsed.

If you want the user to handle backoff IMHO they should handle all retries. This might have some benefits, because the user knows whether or not its safe to retry whereas the client doesn't and is guessing.

If we can instrument the retries so we know how often it happens that would be good, too.

TwP:
@look let me take a shot at adding some backoff to the retry mechanism. Specifying an upper limit on the total request time is a great suggestion. I'll get to that this afternoon.

Adding a `retry_count` to the instrumentation payload should not be too difficult.

look:
LGTM, but I would also recommend a test for the retry behavior. You should be able to do that with WebMock, I think. You can create a mock request and assert that it was requested the expected number of times.

TwP:
@look yup - this definitely needs some testing. I'm very grateful for the feedback here.

My biggest concern is the logic of **only** retrying on `GET` and `HEAD` requests along with only retrying on those three exception classes. Does the logic there hold water?

look:
I think it should be OK to retry any idempotent operation. GET and HEAD are conservative (unless you can think of destructive operations you can issue with those verbs...).

TwP:
Tests have been added. This is ready for a final review and approval.

test/client_test.rb
```diff
@@ -188,4 +197,73 @@
       refute_same $client.connection, client.connection
     end
   end
+
+  describe "automatic retry of requests" do
```
miguelff: :+1: Good coverage!

script/cibuild
```diff
@@ -57,11 +57,12 @@ trap cleanup EXIT
...
 es_version=${ES_VERSION:-24}
-docker_compose="docker-compose --file docker/docker-compose.es${es_version}.yml --file docker/docker-compose.cibuild.yml"
+docker_compose="docker-compose --file docker/docker-compose.es${es_version}.yml"
 
-# if [ "${CI_MODE:-}" ]; then
-#   docker_compose="${docker_compose} --no-ansi"
-# fi
+if [ "${CI_MODE:-}" ]; then
+  docker_compose="${docker_compose} --file docker/docker-compose.cibuild.yml"
+#  docker_compose="${docker_compose} --no-ansi"
```
look: kill this commented out line?

test/client_test.rb
look: I agree, these tests are nice!

script/cibuild
TwP: I'd like to leave it in and uncomment it once we have a more modern version of `docker-compose` out on the build hosts. The `--no-ansi` will suppress the weird control characters in the build output that normally provide colorization in a terminal.

PR #193, opened 2017-12-20
Allow routing for delete_by_query

TwP:
The main object of this pull request is to allow the passing of the `routing` parameter to the application `delete_by_query` implementation. The approach I took was to remove the param validations since the implementation is focused primarily on ES 2.X where parameter validation is not enforced. If ES 5.X or higher is being used, then the native `delete_by_query` implementation is used by default.

### DIFF
--- a/lib/elastomer/client/scroller.rb
+++ b/lib/elastomer/client/scroller.rb
@@ -151,7 +151,7 @@ module Elastomer
       # and https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-type.html#scan
       #
       # client - Elastomer::Client used for HTTP requests to the server
-      # query  - The query to scan as a Hash or a JSON encoded String
+      # query  - The query to scroll as a Hash or a JSON encoded String
       # opts   - Options Hash
       #   :index       - the name of the index to search
       #   :type        - the document type to search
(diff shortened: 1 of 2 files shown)


lib/elastomer/client/scroller.rb
```diff
@@ -135,11 +135,11 @@ def add_sort_by_doc(query)
...
 
-  DEFAULT_OPTS = {
-      :index => nil,
-      :type => nil,
-      :scroll => "5m",
-      :size => 50,
+    DEFAULT_OPTS = {
+      index:   nil,
+      type:    nil,
+      scroll:  "5m",
```
look: This might have something to do with our problem. 5 minutes is a pretty long time. Does the default need to be this big? It is `1m` in the docs.

OTOH, I don't think this changed so it's probably not necessary to change it...

TwP: I had the very same though, and there is a commit where I changed it to `1m`. The philosophy of "change one thing at a time" came to mind and I reverted the default back to `5m`. But yes, it is rather long and we should investigate changing it.

elireisman: Yeah I agree we should lower it to `1m` as a follow up

PR #194, opened 2017-12-20
Always clear the scroll context

TwP:
The change here ensures that the scroll context is cleared when no more results are returned from the scroller. This applies to `scan` queries which are implemented via the `Scroller`.


elireisman:
Oh yeah nice! I added this at $lastjob as a precaution because we allowed customers to consume our data feed API backed by this in small time windows and clearing the context after the last page of each scroll was consumed helped

PR #195, opened 2017-12-28
ES5 compatibility: strip :read_timeout param in native delete-by-query

elireisman:
As stated in title. Fixes some fails in `github-es56` CI build when native delete_by_query is used.

### DIFF
--- a/lib/elastomer/client/native_delete_by_query.rb
+++ b/lib/elastomer/client/native_delete_by_query.rb
@@ -38,6 +38,9 @@ module Elastomer
           raise IncompatibleVersionException, "Elasticsearch '#{client.version}' does not support _delete_by_query"
         end
 
+        # COMPATIBILITY: read_timeout param is invalid, must remove at last moment
+        parameters.select! { |k, _| k.to_sym != :read_timeout }
+
         parameters.keys.each do |key|
           unless PARAMETERS.include?(key) || PARAMETERS.include?(key.to_sym)
             raise IllegalArgument, "'#{key}' is not a valid _delete_by_query parameter"
@@ -47,7 +50,6 @@ module Elastomer
         @client = client
         @query = query
         @parameters = parameters
-
       end
 
       def execute
--- a/lib/elastomer/version.rb
+++ b/lib/elastomer/version.rb
@@ -1,5 +1,5 @@
 module Elastomer
-  VERSION = "3.0.1"
+  VERSION = "3.0.2"
 
   def self.version
     VERSION
--- a/test/client/native_delete_by_query_test.rb
+++ b/test/client/native_delete_by_query_test.rb
@@ -72,6 +72,11 @@ describe Elastomer::Client::NativeDeleteByQuery do
         end
       end
 
+      it "removes :read_timeout param from native delete-by-query calls, since ES5 removes it" do
+        dbq = Elastomer::Client::NativeDeleteByQuery.new(@index.client, {}, read_timeout: "2m")
+        assert_equal dbq.parameters.key?(:read_timeout) || dbq.parameters.key?("read_timeout"), false
+      end
+
       it "deletes by query when routing is specified" do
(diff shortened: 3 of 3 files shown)


elireisman:
note to self: add more client side changes to bundle for one push to github/github, or if this is all, don't forget to [re-release the gem](https://rubygems.org/gems/elastomer-client)

look:
Oh yeah, a test would be nice too.

elireisman:
@look thanks for taking a look at this - yeah I was on the fence, I could definitely be convinced to add this to the VersionSupport lists so the client complains suitably and then try to handle it on the caller side in instead. Unsure if there will be similar cases we should be ready to support yet but not seeing any in the ES5 CI test build yet.

I'm _kinda_ OK with the current in-the-client approach only because the native method is only available on ES5 anyway, but maybe we should emit a warning or something so the end user will know to deal with this too?

elireisman:
still looking at this to see if we can do it a better way on the calling side. Since this is a request param set up in the client configs by default either way, fixing it here seems better. Noticed [this](https://github.com/github/elastomer-client/blame/master/lib/elastomer/client.rb#L196) in the client that makes me think this shouldn't be an issue anymore unless it's being re-applied just for native delete by q, which I don't think it is, so I'm looking into that and making an updated ES5 test branch to run in CI for gh/gh and see if the current approach is still required or if we can do this better another way if so. more TBA!

lib/elastomer/version.rb
```diff
@@ -1,5 +1,5 @@
 module Elastomer
-  VERSION = "3.0.1"
+  VERSION = "3.0.2"
```
look: I think this is generally better to do in a separate commit, so we can bump the version for a few changes at once. 

You probably had to in order to publish it locally but I wouldn't check it in.

test/client/native_delete_by_query_test.rb
```diff
@@ -72,6 +72,11 @@
         end
       end
 
+      it "removes :read_timeout param from native delete-by-query calls, since ES5 removes it" do
+        dbq = Elastomer::Client::NativeDeleteByQuery.new(@index.client, {}, read_timeout: "2m")
+        assert_equal dbq.parameters.key?(:read_timeout) || dbq.parameters.key?("read_timeout"), false
```
look: IMHO the arguments should be reversed, because expected is supposed to go first: http://docs.seattlerb.org/minitest/Minitest/Assertions.html#method-i-assert_equal

lib/elastomer/client/native_delete_by_query.rb
```diff
@@ -38,6 +38,9 @@ def initialize(client, query, parameters)
           raise IncompatibleVersionException, "Elasticsearch '#{client.version}' does not support _delete_by_query"
         end
 
+        # COMPATIBILITY: read_timeout param is invalid, must remove at last moment
+        parameters.select! { |k, _| k.to_sym != :read_timeout }
```
look: As mentioned in chat, `#reject!` is more idiomatic.

test/client/native_delete_by_query_test.rb
elireisman: ah thanks, will update this

lib/elastomer/client/native_delete_by_query.rb
elireisman: same here

elireisman:
update: probably going to push an alternate PR branch for fixing this in github/github but having done, that I'm a bit on the fence as we have hit several other bad params we'll need to deal with besides this one.

I like the idea of the client checking params on those across versions and raising as it is doing for some params now, just for consistency and to let end users handle this as they choose, but then us handling all of these in the github/github code if we can. I might roll a few of those into one PR to save time if doing it in gh/gh isn't messy, more TBA on this soon...

elireisman:
probably just going to handle this on the caller side see also: https://github.com/github/github/pull/83459 will close this issue if we go that way

PR #196, opened 2018-01-03
Handle op_type compatibility in VersionSupport

elireisman:
as title suggests. Shims the param lists we use for validity checks across versions, also provides a shim method for use by callers during upgrade transition.

### DIFF
--- a/lib/elastomer/version_support.rb
+++ b/lib/elastomer/version_support.rb
@@ -13,7 +13,7 @@ module Elastomer
       parent
       refresh
     ].freeze
-    ES_2_X_INDEXING_PARAMETER_NAMES = %i[consistency ttl timestamp].freeze
+    ES_2_X_INDEXING_PARAMETER_NAMES = %i[type consistency ttl timestamp].freeze
     ES_5_X_INDEXING_PARAMETER_NAMES = %i[wait_for_active_shards].freeze
     KNOWN_INDEXING_PARAMETER_NAMES =
       (COMMON_INDEXING_PARAMETER_NAMES + ES_2_X_INDEXING_PARAMETER_NAMES + ES_5_X_INDEXING_PARAMETER_NAMES).freeze
@@ -78,7 +78,20 @@ module Elastomer
         {enabled: b}
       else
         b
-      end 
+      end
+    end
+
+    # COMPATIBILITY: handle _op_type -> op_type request param conversion for put-if-absent bnehavior
+    # Returns the (possibly mutated) params hash supplied by the caller.
+    #
+    # https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#operation-type
+    def op_type(params = {})
+      if es_version_5_x? && params.key?(:_op_type)
+        v = params[:_op_type]
+        params.delete(:_op_type)
+        params[:op_type] = v
+      end
+      params
     end
 
     # Elasticsearch 2.0 changed some request formats in a non-backward-compatible
@@ -136,6 +149,7 @@ module Elastomer
       @indexing_directives = indexing_parameter_names.each_with_object({}) do |key, h|
         h[key] = "_#{key}"
       end
+      check_op_type(@indexing_directives) # hack: valid param loses underscore between ES 2.x & 5.x
       @indexing_directives.freeze
     end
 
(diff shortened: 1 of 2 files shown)


elireisman:
wrt `elastomer-client-es56` build, looks like `docker_default` network setup issue? https://janky.githubapp.com/9021098/output this works on ES 5.6 tests locally

lib/elastomer/version_support.rb
```diff
@@ -78,7 +78,20 @@ def strict_boolean(b)
...
-      end 
+      end
+    end
+
+    # COMPATIBILITY: handle _op_type -> op_type request param conversion for put-if-absent bnehavior
+    # Returns the (possibly mutated) params hash supplied by the caller.
+    #
+    # https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#operation-type
+    def op_type(params = {})
+      if es_version_5_x? && params.key?(:_op_type)
```
look: You might need to handle string keys as well I think.

lib/elastomer/version_support.rb
```diff
@@ -78,7 +78,20 @@ def strict_boolean(b)
...
+        v = params[:_op_type]
+        params.delete(:_op_type)
+        params[:op_type] = v
```
look: You can write this as `params[:op_type] = params.delete(:_op_type)`

lib/elastomer/version_support.rb
```diff
@@ -151,9 +165,18 @@ def unsupported_indexing_directives
...
       end
+      check_op_type(@unsupported_indexing_directives) # hack: valid param loses underscore between ES 2.x & 5.x
       @unsupported_indexing_directives.freeze
     end
 
+    # COMPATIBILITY
+    # If _op_type param is found in map of symbols to request param names, check for ES 5.x and shim
+    def check_op_type(params = {})
+      if es_version_5_x? && params.key?(:op_type)
+        params[:op_type] = "op_type"
```
look: Isn't this going to overwrite whatever was there with `"op_type"`?

elireisman:
note to self: also will need to bump the Gem version accordingly before pushing this (I have another PR out for client changes I could bundle the push with that does this, or bump again if not)

lib/elastomer/version_support.rb
```diff
@@ -151,9 +165,18 @@ def unsupported_indexing_directives
       @unsupported_indexing_directives = unsupported_keys.each_with_object({}) do |key, h|
         h[key] = "_#{key}"
       end
+      check_op_type(@unsupported_indexing_directives) # hack: valid param loses underscore between ES 2.x & 5.x
       @unsupported_indexing_directives.freeze
     end
 
+    # COMPATIBILITY
+    # If _op_type param is found in map of symbols to request param names, check for ES 5.x and shim
+    def check_op_type(params = {})
```
look: I would maybe call this `fix_op_type` or something that indicates it mutates the input. Some Rubyists would also add a `!` on the end of the method to indicate this, not sure if we do here.

lib/elastomer/version_support.rb
```diff
@@ -13,7 +13,7 @@ class VersionSupport
       parent
       refresh
     ].freeze
-    ES_2_X_INDEXING_PARAMETER_NAMES = %i[consistency ttl timestamp].freeze
+    ES_2_X_INDEXING_PARAMETER_NAMES = %i[type consistency ttl timestamp].freeze
```
look: Is this supposed to be `op_type` or is it an unrelated change?

lib/elastomer/version_support.rb
elireisman: Yeah the check is a hack to maintain those internal supported/unsupported lists in VersionSupport, and then `op_type` will be the shim for things like `def params` in the `audit_log` in gh/gh

lib/elastomer/version_support.rb
elireisman: oops thats a typo from a more invasive approach I rejected, will remove thanks!!

lib/elastomer/version_support.rb
elireisman: In the instances I saw the param set in gh/gh it looked to be symbols only there but I can update for safety, good call. I have seen symbol and string keys used on the caller side for some of the index mappings too (the mapping shim patch fixes some of these)

lib/elastomer/version_support.rb
elireisman: ah good call thank you!

lib/elastomer/version_support.rb
elireisman: yeah in the case of the `check_op_type` soon to be `fix_op_type`, this is to maintain the accepted/bad param lists across the versions we're building there. those symbol lists all get mapped to dictionaries of `:key` = `"_key"` for convenience but this one breaks the mold :(

look: Oh right! I remember that `:key => "_key"` business. Carry on :)

PR #197, opened 2018-01-04
alias for strict request params by ES version

elireisman:
for caller-side shims removing illegal unexpected req params when talking to ES5+ needed for `:read_timeout` for `native_delete_by_query`, `:timeout` for `_count` and `_refresh` params, and others (gh/gh PR for this on the way once this is checked in)

### DIFF
--- a/lib/elastomer/version_support.rb
+++ b/lib/elastomer/version_support.rb
@@ -135,6 +135,11 @@ module Elastomer
     end
 
     # COMPATIBILITY
+    #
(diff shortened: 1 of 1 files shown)


elireisman:
thanks, looking into it with dev-exp now!

elireisman:
no answer yet from dev-exp folks but after my other gh/gh deploy I reran this and this time it went through, so guessing the fails are temp anyway. pushing this out & closing

PR #198, opened 2018-01-11
Request Parameter Validation

TwP:
This PR implements validation of the request parameters before sending requests to the Elasticsearch REST API endpoints. Specifically, it ensures that we only send parameters that Elasticsearch is expecting.

There are two parts to this implementation.

### REST API Spec Generator

Elasticsearch declares a [REST API Spec](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/api) which is a collection of JSON files describing the URLs and parameters that Elasticsearch understands. This PR introduces the [`script/generate-rest-api-spec`](https://github.com/github/elastomer-client/blob/parameter-validation/script/generate-rest-api-spec) script which downloads these JSON documents and parses them into Ruby classes usable by `elastomer-client`.

The generator script produces an [`ApiSpec`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client/rest_api_spec/api_spec.rb) class that contains descriptions of all the [`RestApi`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client/rest_api_spec/rest_api.rb) endpoints. The versions generated thus far are:

* [`ApiSpecV2_3`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client/rest_api_spec/api_spec_v2_3.rb)
* [`ApiSpecV2_4`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client/rest_api_spec/api_spec_v2_4.rb)
* [`ApiSpecV5_6`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client/rest_api_spec/api_spec_v5_6.rb)

The specific API spec version is lazily loaded by the [`Elastomer::Client#api_spec`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client.rb#L102) method. 

### Selecting Params

Each invocation of the [`Elastomer::Client#request`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client.rb#L204) method now passes along a `:rest_api` identifier. This identifier is used to lookup the appropriate `RestApi` descriptor from the `ApiSpec`. The request parameters are filtered according the API spec.

You can find this filtering in the [`Elastomer::Client#expand_path`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/client.rb#L343) method.

### Remaining Tasks

What remains now is to add this `:rest_api` identifier to **all** the client requests made in the `elastomer-client` gem. This is a bit of detail work and it should be done in a day.

With this parameter validation in place, we can revisit the logic implemented in the [`VersionSupport`](https://github.com/github/elastomer-client/blob/parameter-validation/lib/elastomer/version_support.rb) class.

/cc @elireisman && @look

### DIFF
--- a/lib/elastomer/client/rest_api_spec.rb
+++ b/lib/elastomer/client/rest_api_spec.rb
@@ -24,7 +24,7 @@ module Elastomer
       # Internal: Load the specific ApiSpec version class for the given version.
       def self.load_api_spec(version)
         path = File.expand_path("../rest_api_spec/api_spec_v#{to_class_version(version)}.rb", __FILE__)
-        if File.exists? path
+        if File.exist? path
           load path
         else
           raise RuntimeError, "Unsupported REST API spec version: #{version}"
--- a/lib/elastomer/client/rest_api_spec.rb
+++ b/lib/elastomer/client/rest_api_spec.rb
@@ -6,3 +6,6 @@ module Elastomer
     end
   end
 end
+
+require_relative "rest_api_spec/api_spec"
+require_relative "rest_api_spec/rest_api"
--- a/lib/elastomer/client/rest_api_spec.rb
+++ b/lib/elastomer/client/rest_api_spec.rb
@@ -2,9 +2,15 @@
 module Elastomer
   class Client
 
-    # Provides access to the versioned RES API specs for Elasticsearch.
+    # Provides access to the versioned REST API specs for Elasticsearch.
     module RestApiSpec
 
+      # Returns an ApiSpec instance for the given Elasticsearcion version. This
+      # method will load the ApiSpec version class if it has not already been
+      # defined. This prevents bloat by only loading the version specs that are
+      # needed.
+      #
+      # Because of this lazy loading, this method is _not_ thread safe.
       #
       # version - the Elasticsearch version String
       #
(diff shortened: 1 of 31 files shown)


lib/elastomer/client/rest_api_spec.rb
```diff
@@ -0,0 +1,40 @@
+
+module Elastomer
+  class Client
+
+    # Provides access to the versioned RES API specs for Elasticsearch.
```
elireisman: typo: missing T in _REST_

elireisman:
one other thing we might do for end users is add a README section (or a sub README `script/`?) detailing pulling down, generating, and checking in an new spec?

elireisman:
👀 the new changes, LGTM!

TwP:
Turns out that the `timeout` parameter is missing from the REST API spec JSON files for the `tasks.list` and `tasks.get` API endpoints: https://github.com/elastic/elasticsearch/issues/28200

I'm going to manually add the `timeout` param to our generated `api_specV5_6.rb` ruby classes (just using 5.6 as an example).

PR #199, opened 2018-01-13
Parameter validation cleanup

TwP:
This removes parameter validation code from the `Docs` module and relies on the `ApiSpec` validations that we now have in the client request method.

/cc @look && @elireisman

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -15,7 +15,6 @@ module Elastomer
       Docs.new self, name, type
     end
 
-
     class Docs
       # Create a new document client for making API requests that pertain to
       # the indexing and searching of documents in a search index.
@@ -554,6 +553,11 @@ Percolate
         client.multi_percolate(params, &block)
       end
 
+      SPECIAL_KEYS= %i[
+        index type id version version_type op_type routing parent timestamp ttl
+        consistency replication refresh wait_for_active_shards
+      ].inject({}) { |h,k| h[k] = "_#{k}"; h }.freeze
+
       # Internal: Given a `document` generate an options hash that will
       # override parameters based on the content of the document. The document
       # will be returned as the value of the :body key.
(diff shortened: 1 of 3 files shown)


lib/elastomer/client/docs.rb
```diff
@@ -554,6 +553,11 @@ def multi_percolate(params = {}, &block)
         client.multi_percolate(params, &block)
       end
 
+      SPECIAL_KEYS= %i[
```
elireisman: There's one corner case to consider where the param remains but no longer has the leading underscore on ES5 such as `_op_type` ➡️ `op_type` that I'm handling in the monkeypatch shims in github/github right now, but we could conceivably handle in the client too/instead if you want?

We'd probably still want the client to fail fast for mismatches and let the end user handle as they see fit instead of silently shimming as we do for our own upgrade.

TwP: When I was first writing the `elastomer-client` gem I noticed that Elasticsearch would return extra information along with documents when you did a search. It returns things like `_version`, `_index`, `_type`, `_id`, etc. I thought it would be clever to use this same style of underscore prefix to denote request params that can be passed to Elasticsearch when indexing - a convenient shorthand.

All the leading underscores are removed when these special keys are converted to request params. Those request params then go through the normal parameter validation flow that we implemented this past week.

So I'm not entirely sure what the special `op_type` case is - this code will **always** remove the leading underscore.

elireisman: Ah I think I see what's up with `op_type` - ES5 build was failing on bad `_op_type` param for audity entry and I shimmed it [here](https://github.com/github/github/blob/0916b7724e5fe25a29d4d39eb6753f1914a64a7c/lib/elastomer/adapters/audit_entry.rb#L50), but looks like it was just being passed wrong with the underscore on ES2 also and getting by silently. I'll try it out and if so, remove the shim now.

PR #200, opened 2018-01-16
Removing `op_type` shims

TwP:
This removes the `op_type` shim methods from the `VersionSupport` class. They are no longer needed now that we have parameter validation built into the client.

/cc @elireisman

### DIFF
--- a/lib/elastomer/version_support.rb
+++ b/lib/elastomer/version_support.rb
@@ -78,17 +78,6 @@ module Elastomer
       end
     end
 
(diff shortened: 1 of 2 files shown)


elireisman:
Prereq: before we upgrade the gem in `github/github` with this checked in, I'll need to deploy [this](https://github.com/github/github/pull/83896)

ndonewar:
Similar work done as part of https://github.com/github/elastomer-client/pull/250

PR #204, opened 2018-04-20
Make the Ruby version changeable on cibuild

ganta:
To change Ruby version and take a test easily in local environment.


ndonewar:
Thanks for your contribution! I'm closing this PR though since it targets a single file that no longer exists (the CI system has changed).

PR #208, opened 2018-12-20
Add custom non-fatal error for rejected execution exceptions

kevinsawicki:
This pull requests adds a custom `RequestError` subclass for `es_rejected_execution_exception` exceptions.

This error is marked as non-fatal.

* https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/common/util/concurrent/EsRejectedExecutionException.java
* https://aws.amazon.com/premiumsupport/knowledge-center/resolve-429-error-es/

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -407,6 +407,7 @@ module Elastomer
         case root_cause["type"]
         when "index_not_found_exception"; raise IndexNotFoundError, response
         when "illegal_argument_exception"; raise IllegalArgument, response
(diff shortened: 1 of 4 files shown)


look:
👍 I think it makes sense to treat this equivalent to a timeout because the server did not process the message. There is always concurrency concerns with resending a message, but applications will need to deal with that (for example, by fetching fresh data when retrying)

yassskatie:
@look What level of version bump would we consider this kind of change?

look:
I'd probably do a patch release for this, I guess? It's not an API change.

PR #210, opened 2019-06-26
Bump Elastomer-Client to v3.1.5

elireisman:
Hi! this is to bump the version and add a 3.1.5 tag as a prereq to releasing the new RubyGems version 🎉 

I am still battling some problems with my local env that I'm going to ask about in #friction (feel free to drop ideas here if this rings a bell) before I push the release, as I can't even run the test suite locally right now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.1.5 (2019-06-26)
+- Add new more granular exception type
+
 ## 3.1.1 (2018-02-24)
 - Output opaque ID information when a conflict is detected
 - Updating the `semantic` gem
--- a/lib/elastomer/version.rb
+++ b/lib/elastomer/version.rb
@@ -1,5 +1,5 @@
 module Elastomer
-  VERSION = "3.1.4"
+  VERSION = "3.1.5"
 
   def self.version
     VERSION


elireisman:
Details on the local env problems I'm having trying to build, test, and release this Gem are below. Disclaimer: I'm a novice in the details of Ruby/Rails build env best practices, so some of this might be silly.

1. I have tried via `rubyenv` to run my local elastomer-client checkout on Ruby `2.3.4`,  2.3.3`, and `2.4.1`
2. I have rehashed the local env and run `bundle clean --force` and then `bundle install` for each version change to start as fresh as possible
3. I did a release a year ago from this same checkout on Ruby `2.4.1` and IIRC it went flawlessly and I haven't touched this checkout since then
4. all Rake tasks I try to run, including `test` and (more important here) `release` ends like so:
```
$ rake test
    Sorry, you can't use byebug without Readline. To solve this, you need to
    rebuild Ruby with Readline support. If using Ubuntu, try `sudo apt-get
    install libreadline-dev` and then reinstall your Ruby.

File does not exist:

rake aborted!
Command failed with status (1)

Tasks: TOP => test
(See full trace by running task with --trace)
```
....and yet a Homebrew-based `readlne` is already in place on my machine, and my `brew upgrade readline` responds `Error: readline 8.0.0_1 already installed`.

Further complicating things, attempts to reinstall Ruby `2.3.3` and `2.3.4` (which seems to have worked for the last-known-good release by @kevinsawicki ) seem to hang forever like so:
```
$ rbenv install 2.3.3
rbenv: /Users/elireisman/.rbenv/versions/2.3.3 already exists
continue with installation? (y/N) y
ruby-build: use openssl from homebrew
Downloading ruby-2.3.3.tar.bz2...
-> https://cache.ruby-lang.org/pub/ruby/2.3/ruby-2.3.3.tar.bz2
Installing ruby-2.3.3...
ruby-build: use readline from homebrew
```

elireisman:
Thanks again @kevinsawicki for the assist on my build problems, much appreciated ✨💖✨

PR #220, opened 2019-11-25
Line-45-46

jymok12:
Index has methods to manage clusters.
More technical this way

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,7 @@ client.cluster.templates
 
 #### Index
 
-The methods in the [Index](lib/elastomer/client/index.rb) class deal with the
-management of indexes in the cluster. This includes setting up type mappings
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,8 +42,7 @@ client.cluster.templates
 
 #### Index
 
-The methods in the [Index](lib/elastomer/client/index.rb) class deal with the
-management of indexes in the cluster. This includes setting up type mappings
+Index has methods to manage clusters.This includes setting up type mappings
```
chrisbloom7: Thank you for your contribution to the project :) Could you please elaborate on why you feel this change improves the documentation? I think it's useful to maintain a link to the class being described. You will also need to fix the space that was removed between the two sentences.

PR #221, opened 2019-11-25
Line 20-21

jymok12:
The wording has been changed to make the sentence sound simpler

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ $ bundle exec rake test
 
 The client provides a one-to-one mapping to the Elasticsearch [API
 endpoints](https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html).
-The API is decomposed into logical sections and accessed according to what you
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,8 +17,7 @@ $ bundle exec rake test
 
 The client provides a one-to-one mapping to the Elasticsearch [API
 endpoints](https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html).
-The API is decomposed into logical sections and accessed according to what you
-are trying to accomplish. Each logical section is represented as a [client
+Logical sections of APIs is used to access the accomplishment. Each logical section is represented as a [client
```
chrisbloom7: Thanks for your second contribution to this project. I agree that the original sentence is a bit complex and could be simplified, but I think in this case it might be more helpful to future readers to expand on the existing text instead of shorten it, perhaps simplifying the concepts in the process. Would you like to take a crack at that type of approach and see if you can make it more readable?

PR #225, opened 2020-02-25
Use kwargs explicitly in call to Url.new

kytrinyx:
In Ruby 2.7 passing a Hash to a method that is expecting keyword arguments triggers
a warning.

The `url` value is a Hash here.


kytrinyx:
> It's surprising to me that a keyword named `url` is expected to be a hash of url components

This was surprising to me as well.

Thanks for the quick review!

PR #229, opened 2022-11-02
Additional Ruby versions in CI workflow

djdefi:
Adds matrix coverage for Ruby 2.3-2.6 -- Adding newer Rubies begins causing failures.

Supersedes and closes https://github.com/github/elastomer-client/pull/201


elireisman (approved):

djdefi:
Just realized there is an error in the matrix setup, as is this is not testing both series of Elasticsearch. 

Will take a look at this again before merging.

djdefi:
That looks better. Is now properly testing each Elasticsearch version against each Ruby version configured in the workflow.

elireisman (approved):
😍

PR #231, opened 2022-12-08
Change port to use default of 9200

ndonewar:
Change the port used to run ES locally from 19200 to 9200, which is the default. That makes it easier to run tests and develop locally without having to override the port (e.g., with `ES_PORT`).

**Notes:**
- change in the Docker files
- remove the client.md section on Boxen (no longer used) which talks about running it on 19200
- update anywhere 19200 was used in docs

### DIFF
--- a/docker/docker-compose.es24.yml
+++ b/docker/docker-compose.es24.yml
@@ -23,7 +23,7 @@ services:
       - esrepos:/usr/share/elasticsearch/repos
       - ./elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml
     ports:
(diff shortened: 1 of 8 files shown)


richa-d (approved):

look:
FWIW, the reason why this used a non-default port was so you could run a the normal version of Elasticsearch on your machine at the same time.

ndonewar:
> FWIW, the reason why this used a non-default port was so you could run a the normal version of Elasticsearch on your machine at the same time.

Makes sense! We're going to be doing some more docker reorganization soon to help with local dev and running multiple instances, so we'll keep this in mind.

PR #236, opened 2022-12-20
Fix cluster_test, multi_search_test and time_test

richa-d:
Changes to ensure tests pass when run on ES 7 for:
- https://github.com/github/elastomer-client/blob/main/test/client/cluster_test.rb
- https://github.com/github/elastomer-client/blob/main/test/client/multi_search_test.rb
- https://github.com/github/elastomer-client/blob/main/test/core_ext/time_test.rb

### DIFF
--- a/test/client/cluster_test.rb
+++ b/test/client/cluster_test.rb
@@ -62,24 +62,24 @@ describe Elastomer::Client::Cluster do
   end
 
   it "updates the cluster settings" do
-    @cluster.update_settings transient: { "indices.ttl.interval" => "30m" }
+    @cluster.update_settings transient: { "indices.recovery.max_bytes_per_sec" => "30mb" }
     h = @cluster.settings
 
-    value = h["transient"]["indices"]["ttl"]["interval"]
+    value = h["transient"]["indices"]["recovery"]["max_bytes_per_sec"]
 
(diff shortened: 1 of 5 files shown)


test/client/cluster_test.rb
```diff
@@ -62,24 +62,24 @@
   end
 
   it "updates the cluster settings" do
-    @cluster.update_settings transient: { "indices.ttl.interval" => "30m" }
+    @cluster.update_settings transient: { "indices.recovery.max_bytes_per_sec" => "30mb" }
```
richa-d, line 65: The TTL setting is not valid in ES 7 so I updated the test to use a different setting.

test/test_helper.rb
```diff
@@ -300,6 +300,7 @@ def mappings_wrapper(type, body)
       }
     }
     mapping[type] = body
+    if disable_all then mapping[type]["_all"] = { "enabled": false } end
```
richa-d, line 303: `_all: { enabled: false }` is not valid from ES 7 - https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#all-meta-field-removed

ndonewar (approved):
LGTM!

PR #237, opened 2022-12-21
Fix index_test.rb for ES 7

richa-d:
Found a few breaking changes here:
- `_percolate` and `_mpercolate` endpoints removed: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#breaking_60_percolator_changes
- `_suggest` endpoint removed: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-suggesters.html
I also added a test for the term suggester to multi_search_test.rb using [this documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-suggesters.html) for reference.
- Using wildcards to get aliases does not raise an error if no match is found. I did not find any documentation for this, but I found documentation on similar behavior for other endpoints - https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#_open_close_index_api_allows_wildcard_expressions_that_match_no_indices_by_default

### DIFF
--- a/test/client/index_test.rb
+++ b/test/client/index_test.rb
@@ -155,7 +155,7 @@ describe Elastomer::Client::Index do
         author: $client.version_support.keyword
       }}}
     end
-    
+
     assert_property_exists @index.mapping[@name], "book", "author"
     assert_property_exists @index.mapping[@name], "book", "title"
 
--- a/test/client/index_test.rb
+++ b/test/client/index_test.rb
@@ -54,39 +54,33 @@ describe Elastomer::Client::Index do
     it "adds mappings for document types" do
       @index.create(
         settings: { number_of_shards: 1, number_of_replicas: 0 },
-        mappings: {
-          doco: {
-            _source: { enabled: false },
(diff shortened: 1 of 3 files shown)


test/client/index_test.rb
```diff
@@ -381,79 +388,87 @@
       assert_equal id, percolator.id
     end
 
-    it "performs multi percolate queries" do
-      # COMPATIBILITY
-      if requires_percolator_mapping?
-        @index.update_mapping("percolator", { properties: { query: { type: "percolator" } } })
-      end
+    it "performs multi percolate queriess" do
```
ndonewar, line 391:
```suggestion
    it "performs multi percolate queries" do
```
Minor: removing an extra `s`

richa-d: Ah sorry I missed that!

ndonewar (approved):
Thanks for adding the test! LGTM!

PR #238, opened 2022-12-22
[ES7 upgrade] Fix docs tests and skip percolator and multi percolator tests

richa-d:
This PR fixes tests for docs on ES 7.
I also added changes to skip percolator and multi percolator tests for ES 7, since those endpoints were removed ES 6 onwards and adding new tests will need more changes, and we don't use the percolator endpoints on dotcom.

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -225,8 +225,11 @@ module Elastomer
       # See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-shards.html
       #
       # Returns the response body as a hash
-      def search_shards(params = {})
-        response = client.get "/{index}{/type}/_search_shards", update_params(params, action: "docs.search_shards", rest_api: "search_shards")
+      def search_shards(params = {}, remove_type_param = false)
+        updated_params = update_params(params, action: "docs.search_shards", rest_api: "search_shards")
+        if remove_type_param then updated_params.delete(:type) end
+
+        response = client.get "/{index}{/type}/_search_shards", updated_params
         response.body
       end
 
--- a/test/client/docs_test.rb
+++ b/test/client/docs_test.rb
@@ -11,25 +11,16 @@ describe Elastomer::Client::Docs do
     unless @index.exists?
       @index.create \
         settings: { "index.number_of_shards" => 1, "index.number_of_replicas" => 0 },
-        mappings: {
-          doc1: {
-            _source: { enabled: true }, _all: { enabled: false },
-            properties: {
-              title: $client.version_support.text(analyzer: "standard", term_vector: "with_positions_offsets"),
-              author: $client.version_support.keyword
-            }
-          },
-          doc2: {
-            _source: { enabled: true }, _all: { enabled: false },
-            properties: {
-              title: $client.version_support.text(analyzer: "standard", term_vector: "with_positions_offsets"),
-              author: $client.version_support.keyword
-            }
+        mappings: mappings_wrapper("book", {
+          _source: { enabled: true },
+          properties: {
+            title: $client.version_support.text(analyzer: "standard", term_vector: "with_positions_offsets"),
+            author: $client.version_support.keyword
           }
-        }
+        }, true)
 
       # COMPATIBILITY
-      if requires_percolator_mapping?
+      if !$client.version_support.es_version_7_plus? && requires_percolator_mapping?
         @index.update_mapping("percolator", { properties: { query: { type: "percolator"}}})
       end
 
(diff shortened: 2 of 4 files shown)


lib/elastomer/client/docs.rb
```diff
@@ -225,8 +225,11 @@ def search(query, params = nil)
       # See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-shards.html
       #
       # Returns the response body as a hash
-      def search_shards(params = {})
-        response = client.get "/{index}{/type}/_search_shards", update_params(params, action: "docs.search_shards", rest_api: "search_shards")
+      def search_shards(params = {}, remove_type_param = false)
```
richa-d, line 228: The update_params method adds the type param to each request. For ES 7 it adds the default type `_doc`. For most endpoints, adding `_doc` works fine for ES 7, but for `_search_shards`, `index/type/_search_shards` does not return the right result. So I'm removing the type param only for this method. This is not the best approach, so I'm open to suggestions for a better way to handle this (I couldn't find a usage for this endpoint on dotcom so this change shouldn't actually affect anything)

test/client/multi_percolate_test.rb
```diff
@@ -138,4 +138,4 @@ def populate!
     @index.refresh
   end
   # rubocop:enable Metrics/MethodLength
-end
+end unless $client.version_support.es_version_7_plus?
```
ndonewar, line 141: While this method of conditionally not running the test will work, it doesn't let the file communicate to the test runner that there are tests that have been skipped. For example, running this test against ES 7 outputs the following:
```
0 runs, 0 assertions, 0 failures, 0 errors, 0 skips
```

Alternatively, if a `skip` is conditionally put into the `before` block of the outermost `describe`, then the test runner will know about the skipped tests and output the following:
```
3 runs, 0 assertions, 0 failures, 0 errors, 3 skips
```

Can you make that change here and in `percolator_test.rb`? I see there are other instances of that type of skip in this repo. To be consistent, the logic would be something like the following as the first few lines in the `before` block:

```
if $client.version_support.es_version_7_plus?
  skip "MultiPercolate not supported in ES version #{$client.version}"
end
```

richa-d: Got it, I'll make these changes. Thanks!

richa-d: Done!

test/client/docs_test.rb
```diff
@@ -198,49 +192,49 @@
...
+    h = $client.version_support.es_version_7_plus? ? @docs.get(id: "1") : @docs.get(id: "1", type: "book")
 
     assert_found h
-    assert_equal "mojombo", h["_source"]["author"]
+    assert_equal "Author1", h["_source"]["author"]
   end
 
-  it "checks if documents exist in the search index" do
-    refute @docs.exists?(id: "1", type: "doc1")
+  it "checks if documents exist in the search indexx" do
```
ndonewar, line 207:
```suggestion
  it "checks if documents exist in the search index" do
```
Minor: removed extra `x`

test/client/docs_test.rb
```diff
@@ -582,103 +601,113 @@
...
+    response = $client.version_support.es_version_7_plus? ? @docs.multi_term_vectors({ids: [1, 2]}, fields: "title", term_statistics: true) : @docs.multi_term_vectors({ids: [1, 2]}, type: "book", fields: "title", term_statistics: true)
     docs = response["docs"]
 
     assert docs
     assert_equal(%w[1 2], docs.map { |h| h["_id"] }.sort)
   end
 
   it "percolates a given document" do
-    populate!
+    if !$client.version_support.es_version_7_plus?
```
ndonewar, line 612: Same comment here as the other one regarding using `if/skip/end` rather than wrapping so that the test runner sees a skip. Can you update this one as well?

test/client/docs_test.rb
```diff
@@ -582,103 +601,113 @@
...
-    assert_equal 1, response["matches"].length
-    assert_equal "1", response["matches"][0]["_id"]
+      assert_equal 1, response["matches"].length
+      assert_equal "1", response["matches"][0]["_id"]
+    end
   end
 
   it "percolates an existing document" do
-    populate!
+    if !$client.version_support.es_version_7_plus?
```
ndonewar, line 633: Looks like one more related to the previous comment.

test/client/docs_test.rb
richa-d: Done!

lib/elastomer/client/docs.rb
ndonewar: Based on our discussion today, I think this is the way to go. Even after we handle defaulting the `type` for ES 7 in `update_params`, this one would still get a `type` so removing it here seems good for now.

ndonewar (approved):
Thanks for the changes - LGTM!

PR #239, opened 2023-01-03
[ES 7 upgrade] Fix scroller, snapshot, template, delete by query tests

richa-d:
This PR fixes scroller, snapshot, template, delete by query tests for ES 7. 
Once this PR merges, all tests should pass for ES 7.

Notes:
- While fixing the native_delete_by_query tests, I found that the default `number_of_shards` for an index in ES 7 is 1, while in ES 5 it was 5. Since ES 7 was only creating 1 shard, deleting by specifying `_routing` values did not work because all the documents were in the same shard ([link to documentation for _routing](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/mapping-routing-field.html)). I also noticed that the calculation of the shard number for each document we index has changed in ES 7:

ES 7:
image
ES 5:
image
I am not sure if and how this routing change in ES 7 affects github/github, but thought it was worth noting.

- app_delete_by_query is [deprecated](https://github.com/github/elastomer-client/blob/main/lib/elastomer/client/app_delete_by_query.rb#L6) ES 5 onwards, so I skipped those tests for ES 7
- The cluster setting `ingest.geoip.downloader.enabled` [defaults to true in ES 7](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/geoip-processor.html#geoip-cluster-settings), which was causing snapshot tests to fail, since we assert that the snapshot we create only has 1 index, but we were getting an additional `.geoip_databases` index. So I just updated the cluster setting [here](https://github.com/github/elastomer-client/blob/richa-d/fixTestsPart5/test/client/snapshot_test.rb#L17-L19) to be set to false so that we only have 1 index.
- Changes to bulk.rb, assertions.rb and client.rb are only to fix rubocop errors

### DIFF
--- a/lib/elastomer/client.rb
+++ b/lib/elastomer/client.rb
@@ -16,7 +16,7 @@ module Elastomer
     IVAR_BLACK_LIST = [:@basic_auth, :@token_auth]
     IVAR_NOISY_LIST = [:@api_spec, :@cluster]
 
-    MAX_REQUEST_SIZE = 250 * 2**20  # 250 MB
+    MAX_REQUEST_SIZE = 2**20 * 250  # 250 MB
 
     RETRYABLE_METHODS = %i[get head].freeze
 
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -160,7 +160,7 @@ module Elastomer
     # immediately.
     #
     class Bulk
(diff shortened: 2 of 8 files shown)


ndonewar (approved):
> Once this PR merges, all tests should pass for ES 7.

🎉 LGTM!

PR #240, opened 2023-01-04
Add ES 7 to CI workflow

ndonewar:
With the previous PR, existing tests now pass against ES 7. Add ES 7 to the workflow file to run the full suite against it.


ndonewar:
Closing in favor of #241

PR #241, opened 2023-01-04
Add ES 7 to the CI workflow

ndonewar:
Now that the existing test suite passes against ES 7, add ES 7 to the CI workflow.

Additionally, ES 2.4.6 was removed from the matrix. The `run: "sed...` line was not compatible with ES 7. While a conditional could be added, support for ES 2 is going away soon so it's simpler to remove that line and version.

The workflow also now uses Actions cache for the unzipped ES 2 and ES 5 binaries. This change speeds up the workflow runs and saves bandwidth by skipping the download step. According to the [Actions documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key), feature branches should use the cache of the default branch, which can be verified once this PR is merged. 

Screenshot of skipping the download step:
image

One minor cleanup: removed unused `.overcommit.yml` file.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         ruby-version: ['2.7', '3.0', '3.1']
-        ES_VERSION: ['2.4.6', '5.6.15']
+        ES_VERSION: ['2.4.6', '5.6.15', '7.17.8']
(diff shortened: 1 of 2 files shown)


richa-d (approved):
LGTM! :shipit:

PR #242, opened 2023-01-10
[ES7 upgrade] Make document APIs work even when type is passed for ES7

richa-d:
ES 5 has a type param for all endpoints, while ES 7 does not accept the type param. For ES 7 document APIs, the commonly used endpoints - index, get, delete, exists - need to have an `_doc` in the path. So removing type will not work for document APIs.
Because of this, I decided to add the type param and set it to `_doc`, since this ensures that all the endpoints work as expected. 
To test this change, I added an additional test, which calls all the docs endpoints that we use in dotcom (index, get, update, multi_get, delete, exists, count, search) with a type param.

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -597,6 +597,7 @@ module Elastomer
         h = defaults.update params
         h.update overrides unless overrides.nil?
         h[:routing] = h[:routing].join(",") if h[:routing].is_a?(Array)
+        if client.version_support.es_version_7_plus? then h[:type] = "_doc" end
         h
       end
 
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -597,7 +597,7 @@ module Elastomer
         h = defaults.update params
         h.update overrides unless overrides.nil?
         h[:routing] = h[:routing].join(",") if h[:routing].is_a?(Array)
-        if client.version_support.es_version_7_plus? then h[:type] = "_doc" end
+        h[:type] = "_doc" if client.version_support.es_version_7_plus?
         h
       end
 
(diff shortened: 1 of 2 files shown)


lib/elastomer/client/docs.rb
```diff
@@ -597,6 +597,7 @@ def update_params(params, overrides = nil)
         h = defaults.update params
         h.update overrides unless overrides.nil?
         h[:routing] = h[:routing].join(",") if h[:routing].is_a?(Array)
+        if client.version_support.es_version_7_plus? then h[:type] = "_doc" end
```
ndonewar, line 600: Minor, but I typically see similar Ruby one-liners written this way:
```
h[:type] = "_doc" if client.version_support.es_version_7_plus?
```

test/client/docs_test.rb
```diff
@@ -720,6 +720,60 @@
...
+    assert_equal "2", h["_id"]
+
+    h = @docs.multi_get({ids: [2, 1]}, type: "book")
+    authors = h["docs"].map { |d| d["_source"]["author"] }
+
+    assert_equal %w[Author2 Author1.1], authors
+
+    h = @docs.delete id: 3, type: "book"
+
+    refute @docs.exists?(id: "3", type: "book")
```
ndonewar, line 765: Just checking, did you want to add a doc with an `id` of 3 before these lines? Otherwise the `refute` is slightly misleading since it would pass for (almost) any values of `id` and `type`, even those not matching the `delete` (since `delete` isn't deleting an existing doc).

richa-d: Yup, my bad. The refute line was to ensure that the `delete` worked correctly, and also to test if `.exists?` works with a type.
I moved around a bunch of code and removed the indexing code for id = 3 by mistake. Thanks!

lib/elastomer/client/docs.rb
richa-d: Done, thanks!

ndonewar (approved):
Thanks for the changes! LGTM 🎉

richa-d:
Rubocop is failing on a file that this PR did not change, so I will fix it in my next PR.

PR #243, opened 2023-01-10
[ES7 upgrade] Allow passing type param for Index APIs

richa-d:
For the index APIs, only 3 endpoints have a type param in ES5 - `exists_type`, `get_mapping` and `put_mapping`. These 3 endpoints have an optional type param in ES7, and if you want to pass the type param, you have to also set `include_type_param` to true. `include_type_param` is deprecated in ES7, and removed from ES8. So I thought it would be better to not use the type param at all for ES7, so I removed the type param from all ES7 index API requests. After these changes, if you do pass a type param to ES7, it will be removed before calling the actual index endpoint.

The test I added checks for `get_mapping` and `update_mapping` only. `exists_type` is deprecated in ES7 and removed from 8. We don't expose that endpoint, and I didn't find any usage of this in dotcom, so I did not add a test for it.

### DIFF
--- a/lib/elastomer/client/index.rb
+++ b/lib/elastomer/client/index.rb
@@ -147,7 +147,7 @@ module Elastomer
       #
       # Returns the response body as a Hash
       def update_mapping(type, body, params = {})
-        response = client.put "/{index}/_mapping/{type}", update_params(params, body: body, type: type, action: "index.update_mapping", rest_api: "indices.put_mapping")
+        response = client.put "/{index}/_mapping{/type}", update_params(params, body: body, type: type, action: "index.update_mapping", rest_api: "indices.put_mapping")
         response.body
       end
       alias_method :put_mapping, :update_mapping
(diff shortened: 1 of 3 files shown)


lib/elastomer/client/index.rb
```diff
@@ -147,7 +147,7 @@ def get_mapping(params = {})
       #
       # Returns the response body as a Hash
       def update_mapping(type, body, params = {})
-        response = client.put "/{index}/_mapping/{type}", update_params(params, body: body, type: type, action: "index.update_mapping", rest_api: "indices.put_mapping")
+        response = client.put "/{index}/_mapping{/type}", update_params(params, body: body, type: type, action: "index.update_mapping", rest_api: "indices.put_mapping")
```
ndonewar, line 150: Nice syntax find

ndonewar (approved):
>  I thought it would be better to not use the type param at all for ES7, so I removed the type param from all ES7 index API requests

Good call - LGTM!

PR #245, opened 2023-01-19
[ES upgrade] Add ES 8 support and generate apispec 

richa-d:
This PR adds support for ES 8 (8.6.0).
The tests throw errors for ES 8, I will fix those in a subsequent PR.

**Notes:**

I created another elasticsearch7plus.yml file because there were 2 changes that I found in the cluster settings for ES 7 and 8.
- The `transport.tcp.port` setting was renamed to `transport.port` ES 7 onwards.
- The cluster coordination subsystem that was used in ES 5 (Zen Discovery) was replaced with a different one  - https://www.elastic.co/blog/a-new-era-for-cluster-coordination-in-elasticsearch. Due to this, the `discovery.zen.minimum_master_nodes` setting was being [ignored in 7](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_cluster_bootstrapping_is_required_if_discovery_is_configured) and has been removed from 8 (and was throwing an error while setting up 8), so I removed it from the ES cluster settings for 7 and 8.

### DIFF
--- a/docker/compose.yaml
+++ b/docker/compose.yaml
@@ -1,6 +1,33 @@
 version: "3"
 
 services:
+  elasticsearch8.6:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.6.0
+    container_name: es8.6
+    profiles: ["es8", "all"]
+    environment:
+      - cluster.name=elastomer8.6
+      - bootstrap.memory_lock=true
+      - discovery.type=single-node
+      - xpack.security.enabled=false
+      - xpack.watcher.enabled=false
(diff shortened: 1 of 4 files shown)


ndonewar (approved):
LGTM!

test/version_support_test.rb
```diff
@@ -4,10 +4,11 @@
 
 describe Elastomer::VersionSupport do
   describe "supported versions" do
-    it "allows 2.3.0 to 5.6" do
+    it "allows 2.3.0 to 8.6.0" do
```
davekenn, line 7: Should that be 8.7.0?

PR #246, opened 2023-01-24
[ES upgrade] Fix docs, delete by query tests

richa-d:
This PR fixes tests for docs and delete by query endpoints.
Note: For the docs endpoints, some endpoints like get and delete still need `_doc` types, while the other endpoints do not accept a type.

### DIFF
--- a/test/client/docs_test.rb
+++ b/test/client/docs_test.rb
@@ -721,7 +721,7 @@ describe Elastomer::Client::Docs do
   end
 
   it "accepts a type param and does not throw an error for ES7" do
-    if !$client.version_support.es_version_7_plus?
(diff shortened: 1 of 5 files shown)


test/client/docs_test.rb
```diff
@@ -721,7 +721,7 @@
   end
 
   it "accepts a type param and does not throw an error for ES7" do
-    if !$client.version_support.es_version_7_plus?
+    if !$client.version_support.es_version_7_plus? || $client.version_support.es_version_8_plus?
       skip "This test is only needed for ES 7 onwards"
```
ndonewar, line 725:
```suggestion
      skip "This test is only needed for ES 7"
```

Minor word update since the skip happens for ES 8 now too.

ndonewar (approved):
LGTM!

PR #247, opened 2023-01-24
Fix latest Rubocop warnings about development dependencies

ndonewar:
This PR fixes the latest Rubocop warnings about development dependencies being in the gemspec by moving them to the Gemfile.
```
Gemspec/DevelopmentDependencies: Specify development dependencies in Gemfile.
```
 Two unused gems were removed as cleanup (`awesome_print` and `minitest-fail-fast`), and a few gems were updated to latest versions. `platform: :ruby_19` was removed from `simplecov` since elastomer-client no longer runs against that version of Ruby. As a sanity check, SimpleCov did generate a report when `COVERAGE=true` was specified.


kag728 (approved):
nice - LGTM

PR #249, opened 2023-01-26
Update CI for ES 8

ndonewar:
This PR adds ES 8 to the CI workflow.

Additionally:
- removes Ruby 2.7 - it will be EOL in about 2 months, so keep the CI process smaller
- removes the "Install dependencies (Ubuntu)" step - shortens build time, as updating isn't necessary, and installing a Java 8 JRE isn't needed (or wanted). ES comes bundled with more modern JREs.
- disables HTTPS for ES 8 (and 7). The command line option doesn't work with the ES 5 binary, so there's a conditional for that step.


kag728 (approved):
👍

PR #250, opened 2023-01-30
Remove support for ES 2

ndonewar:
Remove support for ES 2 to minimize the number of maintained versions (currently 5, 7, and 8).

**Notes:**
- bump memory in docker compose file due to ES 7 and 8 running out of memory locally when runnings tests
- for any check that referenced ES 2, replace with whatever would be generated when the check returned `false`
- for any check that referenced ES 5+, replace with whatever would be generated when the check returned `true`
- warmers were completely removed since those were an ES 2 only feature
- Removed all related `app_delete_by_query` since ES 5+ all use `native_delete_by_query`. That code could be refactored to drop "native" from the class names, methods, etc. but that can be done in a subsequent to have a cleaner diff in this one.
- There are a fair number of whitespace changes in the diff, so ignoring those may make the review easier


look:
Ha, this has been a long time coming! 🎉

lib/elastomer/version_support.rb
```diff
@@ -7,104 +7,21 @@ class VersionSupport
 
     attr_reader :version
 
-    # version - an Elasticsearch version string e.g., 2.3.5 or 5.3.0
+    # version - an Elasticsearch version string e.g., 5.6.6 or 7.17.8
     #
     # Raises ArgumentError if version is unsupported.
     def initialize(version)
-      if version < "2.3" || version >= "8.7"
+      if version < "5.0" || version >= "8.7"
```
kag728, line 14: woo!

kag728 (approved):
LGTM

PR #252, opened 2023-02-09
Cleanup version_support.rb

ndonewar:
This PR cleans up the version_support.rb file by removing a few comments and an unused method noticed during a final review of the changes before release. It also bumps the supported version to `< 9.0`. It's reasonable to expect the library to continue working without modification through the 8.x minor releases.


kag728 (approved):
LGTM

PR #254, opened 2023-02-10
Fix version_support check in bulk class

kag728:
Update bulk method to remove `$client` call for `version_support`
`$client` is only available in test methods, and `version_support` is available to call directly on the `bulk` method due to reopening the `Client` class


ndonewar (approved):
LGTM!

Issue #256, opened 2023-03-03
Cannot provide `routing` parameter to Bulk Index API when targeting Elasticsearch 7+


lerebear:
As per [these docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed) a number of underscored parameters (such as `_routing` and `_retry_on_conflict`) have been removed from certain Elasticsearch (ES) APIs as of ES 7.

However, it appears that this client will rewrite non-underscored versions of some of these params to underscored versions before forwarding them to ES:

https://github.com/github/elastomer-client/blob/6ddf4311e60d4228dd390bb9a411a29289eadd6e/lib/elastomer/client/bulk.rb#L314-L326

That prevents consumers of this library from specifying a `routing` parameter to the Bulk Index API of ES 7+, because an attempted use of the API as shown below will fail because the `routing` parameter gets converted to `_routing`:

```ruby
index.bulk do |request|
  request.index({ foo: "bar" }, routing: "123")
end
```

The error I've observed looks like this:

```
Elastomer::Client::IllegalArgument: {"root_cause"=>[{"type"=>"illegal_argument_exception", "reason"=>"Action/metadata line [1] contains an unknown parameter [_routing]"}], "type"=>"illegal_argument_exception", "reason"=>"Action/metadata line [1] contains an unknown parameter [_routing]"}
```

PR #257, opened 2023-03-03
Do not prefix certain params passed to the Bulk API with underscores.

lerebear:
Closes https://github.com/github/elastomer-client/issues/256.

As per [these docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed), as of ES 7 certain params  (e.g. `_routing`) should be passed to the ES Bulk API without the underscore prefix. This ensures that we don't add the underscore prefix to just those params.

## Approach

- I refactored the special key handling to make it a bit easier to follow (previously, the variable names were very terse, and it was difficult to understand the distinction between key and value in `SPECIAL_KEYS_HASH`). Now the logic is driven by two arrays, and we prefix the values of those arrays just-in-time when necessary.
- I stopped adding an underscore prefix to just the keys listed in the documentation above. The result is that you might still receive an `IllegalArgument` error from the API if you use an unsupported underscored param, but you should always get the correct result if you use the correct params.

### DIFF
--- a/lib/elastomer/client/bulk.rb
+++ b/lib/elastomer/client/bulk.rb
@@ -312,7 +312,7 @@ module Elastomer
       end
 
       SPECIAL_KEYS = %w[id type index version version_type routing parent timestamp ttl consistency refresh retry_on_conflict]
-      SPECIAL_KEYS_HASH = SPECIAL_KEYS.inject({}) { |h, k| h[k] = "_#{k}"; h }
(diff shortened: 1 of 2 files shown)


ndonewar (approved):
LGTM!

PR #258, opened 2023-03-03
Pin the Rubocop dependencies

ndonewar:
For now, pin the Rubocop dependencies to patch level updates to keep new PR checks and CI runs from unexpectedly failing.


kag728 (approved):
:shipit:

PR #259, opened 2023-03-04
Bump to version 4.0.2.

lerebear:
This bumps the version of this gem to `4.0.2` in order to package the fix for #256.

I have tested that the new version of the gem works properly in a local environment as per https://github.com/github/elastomer-client#releasing.


ndonewar (approved):
Thanks for adding the 4.0.1 CHANGELOG entry! 🎉

PR #260, opened 2023-03-04
Update scripts and documentation for getting started.

lerebear:
This suggests a few updates to scripts and documentation based on my recent experience getting started with development on this project:

- I've made the bootstrap script a bit more verbose, so that it's easier to understand what it is doing (bd9aaee9).
- ~~I declared a default Ruby version in `.ruby-version`, and committed that file to the repository. I chose the highest Ruby version that the project is tested against, which happens to be the highest stable Ruby version at time of writing (1f54d720).~~ Superseding [PR](https://github.com/github/elastomer-client/pull/288)
- I updated the README with the latest information about compatibility and the release process. I also reformatted the setup instructions slightly to make them easier to copy.


ndonewar (approved):
One change suggested, but otherwise approved!

lerebear:
Thank you for the feedback @ndonewar: I have applied the suggested change.

I also wonder if you have any insight into why the Rubocop build is failing on this branch? It looks like setting the default ruby version to 3.2.1 caused us to pick up a new rule about hash syntax. I'm not sure why that is though, since it appears that previous builds ran on Ruby 3.1.3, which already supported the hash syntax in question. Do you understand what has gone wrong? Alternatively, should I just fix the new warnings?

ndonewar:
@lerebear I don't have any insight offhand into why this one would show up now due to the Ruby change. If you want to fix the warnings, that sounds good. I'm also okay with undoing the Ruby change for now. I'd say it's your choice!

ndonewar (approved):
I moved the RuboCop work out into another [PR](https://github.com/github/elastomer-client/pull/288). The docs and script changes look good!

PR #264, opened 2023-04-06
Add update_by_query

### LINKED ISSUE #263: Add support for `_update_by_query`
It would be good to have support for `_update_by_query` through a supported API. Something like:

```ruby
index.update_by_query(query)
docs.update_by_query(query, _routing: "route")
```

ekroon:
Fixes #263 

This PR adds support for `_update_by_query`. 

I followed the implementation for `_delete_by_query`, but I didn't use the `native_` part. Seemed like some kind of backwards compatibility thing and just makes the interface longer?

### DIFF
--- a/lib/elastomer/client/docs.rb
+++ b/lib/elastomer/client/docs.rb
@@ -289,6 +289,18 @@ module Elastomer
         @client.native_delete_by_query(query, update_params(params))
       end
 
+      # Update documents by query using Elasticsearch's _update_by_query API.
+      #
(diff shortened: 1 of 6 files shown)


ndonewar (approved):
Looks good to me!

> I followed the implementation for _delete_by_query, but I didn't use the native_ part.

Thanks for keeping the implementation consistent with the other one and glad you didn't follow the `native_` part! 😄 That's from a partially completed refactor away from ES 2 support and will eventually be removed.

PR #266, opened 2023-04-07
Bump to version 4.0.3

ekroon:
This bumps the version of the gem to `4.0.3`

There is still an outstanding cleanup in #265, so this should be merged after that gets merged.

I followed the steps in the release section of the README to verify version is updated and new functionality works as expected by recreating a test with manual steps / without helpers.


ndonewar (approved):
🚀

PR #268, opened 2023-04-26
Update bulk API to remove _type if version is > 7

kag728:
This PR intends to fix two different issues: 

1. For ES 7+ compatibility, the Bulk API will no longer send the `_type` parameter if the version is 7+. 

[Link to ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) showing that `_type` is not a parameter for any of the bulk requests

As part of this change, all the bulk tests using this call have been updated to remove the `document_wrapper` helper method to test that the ES 5 requests still work correctly on ES 7+. The Docs API already has been doing this, so I was able to update the docs tests to remove the `document_wrapper` and confirm that all the tests still run properly. 

2. Fixes a small bug where string params were not being removed properly if the parameter was empty or nil. 

I added a test to check the scenario where id is nil or empty, which means that ES should generate its own id hash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.0.1 (2023-04-26)
+- Fix bug in bulk API preventing string `_id` from being removed if empty
+- Remove `_type` from document body during bulk requests for versions ES 7+
 ## 5.0.0 (2023-04-17)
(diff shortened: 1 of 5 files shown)


ndonewar (approved):
LGTM!

PR #269, opened 2023-05-04
Simplify the search_shards check for the type param

ndonewar:
While going through the [ES 6 breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#_search_shards_api), I noticed the following line which suggested that the `type` in the URL path wasn't doing anything in ES 5. That means we can simplify the check for the `search_shards` method by discarding it completely on all supported ES versions.

> The search shards API no longer accepts the type url parameter, which didn’t have any effect in previous versions.


richa-d (approved):
LGTM!

PR #270, opened 2023-05-25
Fix document parameters passed to bulk API when prefixed with underscores

kag728:
Related to https://github.com/github/elastomer-client/pull/257

As per [these docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed), as of ES 7 certain params (e.g. _routing) should be passed to the ES Bulk API without the underscore prefix. The previous PR linked above handled that for parameters explicitly passed to the Bulk request. However, we also pass these parameters with some documents that are then converted to bulk parameters in the `from_document` method.

This PR updates the `from_document` method to handle underscored parameters being passed through the document for ES7+. 

I've added logic to fix this for when the parameter is passed as underscore, and when the parameter is not passed as underscore, so that it will work for both versions of ES5 and in ES8 depending on how you pass the parameter in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.0.3 (2023-05-31)
+- Allow deprecated underscored parameters to work for Bulk API for versions ES 7+. 
+- Allow non-underscored parameters to work for Bulk API for version ES 5. 
 ## 5.0.2 (2023-05-31)
 - Add ES 8.7 REST API spec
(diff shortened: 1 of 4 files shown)


ndonewar (dismissed):
👍

ndonewar (approved):

PR #271, opened 2023-05-30
Add generated api spec for version 8.7

fuentesjr:
It looks like we're missing the spec for 8.7, so I get ahead and generated it via `script/generate-rest-api-spec 8.7`.


ndonewar (approved):
LGTM, thanks!

PR #276, opened 2023-08-07
Reduce the noise of the client during an inspect call by hiding connection info

ndonewar:
Add the `@connection` instance variable to the list of noisy ivars so that it will not be output during calls to `inspect`. The amount of information is quite large and is mostly not helpful during debugging. The information is still available by other means, but removing it from the default output will make debugging clearer.


kag728 (approved):
👍

PR #279, opened 2023-09-14
Update convert_special_keys to properly handle ES8 unprefixed keys

kag728:
This PR handles 2 bugs

1. Adds an additional check to prevent `type` from being used as a special key in ES8. `type` is allowed to be a normal field in ES8, and should not be converted to anything else.
2. As per [these docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed), as of ES 7, certain params (e.g. _routing) should be passed to the ES Bulk API without the underscore prefix. This PR updates `convert_special_keys` method to check for the prefixed keys as well as the non-prefixed keys in the parameters, and converts it to the correct key based on the version of ES.

### DIFF
--- a/lib/elastomer_client/client/bulk.rb
+++ b/lib/elastomer_client/client/bulk.rb
@@ -356,6 +356,8 @@ module ElastomerClient
         opts = {}
 
         SPECIAL_KEYS.each do |key|
+          next if key == "type" && client.version_support.es_version_7_plus?
+
           omit_prefix = (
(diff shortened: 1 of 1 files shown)


ndonewar (approved):
👍

richa-d (approved):

PR #281, opened 2023-09-26
Stop removing unprefixed parameters from document in bulk requests

kag728:
Certain parameters can be sent in the request in the document. There is a method that helps find those parameters, delete them from the document, and extract them as params. 

Currently, elastomer-client allows parameters to be sent either as params or as a document, and then does some manipulation to ensure that the parameters are in the right format for the version of ES that we're working on.

This is complicated, mainly due to different breaking changes that have occurred between ES2, ES5 and ES8 that have different forms of the parameter based on the version. For example, some parameters in ES8 are required to be unprefixed ([see here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_camel_case_and_underscore_parameters_deprecated_in_6_x_have_been_removed)). We shouldn't check for these unprefixed params in ES5 because those can be valid and aren't special keys in ES5.

This PR stops allowing parameters to be sent in with the document and rather puts the burden on the consumer to send parameters as parameters in bulk requests. This reduces the confusion of this method and makes it a lot easier to read the code. 

Other changes include: 
- Bulk test changes to ensure that all parameters are sent separately from the document
- Docs test changes, to change one test to pass in parameters instead of parameters within the document


ndonewar (approved):
LGTM!

PR #283, opened 2023-11-07
Allow a custom Faraday connection configuration block.

misalcedo:
Allow custom configuration of the client's Faraday connection by passing a block to the client's constructor.

Bumped minor version and updated the changelog.


ndonewar (approved):
👍

PR #284, opened 2023-11-30
improvement to update action docs

stephenotalora:
Regretfully is important to note that solely  using the examples documented [here](https://github.com/github/elastomer-client/blob/ec2474b8d1837f4d2cc5fcf5cb839843257d86c6/lib/elastomer_client/client/bulk.rb#L274-L291) leads to an exception:  

**`"x_content_parse_exception"`**

Upon further investigation I came across this [ElasticSearch post](https://discuss.elastic.co/t/updating-elasticsearch-document/265705) with folks having a similar issue.

For future reference I thought it would be best to update these examples  as part of the bulk action code whilst the team has time to update https://github.com/github/elastomer-client/blob/main/docs/bulk_indexing.md

### DIFF
--- a/lib/elastomer_client/client/bulk.rb
+++ b/lib/elastomer_client/client/bulk.rb
@@ -280,9 +280,9 @@ module ElastomerClient
       # params   - Parameters for the update action (as a Hash) (optional)
       #
       # Examples
-      #   update({"doc" => {"foo" => "bar"}}, {:_id => 1}
-      #   update({"doc" => {"foo" => "bar"}}, {:id => 1}
-      #   update({"doc" => {foo" => "bar"}}, "_id" => 1)
+      #   update({"doc" => {"foo" => "bar"}}, {:_id => 1})
+      #   update({"doc" => {"foo" => "bar"}}, {:id => 1})
+      #   update({"doc" => {"foo" => "bar"}}, "_id" => 1)
       #
       # Returns the response from the bulk call if one was made or nil.
       def update(document, params)


lib/elastomer_client/client/bulk.rb
```diff
@@ -280,9 +280,9 @@ def create(document, params)
       # params   - Parameters for the update action (as a Hash) (optional)
       #
       # Examples
-      #   update({"foo" => "bar"}, {:_id => 1}
-      #   update({"foo" => "bar"}, {:id => 1}
-      #   update("foo" => "bar", "_id" => 1)
+      #   update({"doc" => {"foo" => "bar"}}, {:_id => 1}
```
kag728, line 283: Do you know if this is proper syntax for ES5 and ES8? I just want to make sure that if it's only correct for one of them, maybe we keep both and specify which version they each work on

kag728: I double checked through the ES docs and confirmed with myself this is good! Thanks for double checking though 😄

kag728 (dismissed):
🚢

lib/elastomer_client/client/bulk.rb
```diff
@@ -280,9 +280,9 @@ def create(document, params)
...
+      #   update({"doc" => {"foo" => "bar"}}, {:id => 1}
+      #   update({"doc" => {foo" => "bar"}}, "_id" => 1)
```
ndonewar, line 285:
```suggestion
      #   update({"doc" => {"foo" => "bar"}}, {:_id => 1})
      #   update({"doc" => {"foo" => "bar"}}, {:id => 1})
      #   update({"doc" => {"foo" => "bar"}}, "_id" => 1)
```
Minor updates, as all three seem to have slight syntax issues.

stephenotalora:
@kag728 @ndonewar could use another stamp to merge 🙇🏼

lib/elastomer_client/client/bulk.rb
stephenotalora: Thanks for catching these @ndonewar ✨

ndonewar (approved):
👍

PR #286, opened 2023-12-04
Remove default retry logic to force clients to implement on their own

kag728:
This PR removes all default retry logic from elastomer-client, putting the burden on the consumer to implement retry logic using Faraday as they wish. 

This removes `max_retries` and `retry_delay` parameters from the initialization since they're no longer used anywhere. This is a breaking change, and once merged, will require a major version upgrade of elsatomer-client.


test/client_test.rb
```diff
@@ -374,73 +374,4 @@
...
-        to_timeout.then.
-        to_return({
-          headers: {"Content-Type" => "text/plain; charset=UTF-8"},
-          body: "green open test-index 1 0 0 0 159b 159b"
-        })
-
-      assert_raises(ElastomerClient::Client::ConnectionFailed) {
-        $client.get("/_cat/indices")
-      }
-    end
```
misalcedo, line 402: We probably want to keep this test. That way we can keep ensuring it has no retries by default.

test/client_test.rb
```diff
@@ -374,73 +374,4 @@
...
-      stub_request(:get, $client.url+"/test-index/_settings").
-        to_timeout.then.
-        to_timeout.then.
-        to_return({body: %q/{"acknowledged": true}/})
-
-      response = $client.index("test-index").settings(max_retries: 2)
-
-      assert_equal 2, @events.first.payload[:retries]
-      assert_equal({"acknowledged" => true}, response)
-    end
```
misalcedo, line 414: If not too much work, I think it would be good to turn this test into an example of how to add retries to a client using the retry middleware in Faraday. That would require a different client instance since this one is configured without retries already.

What do you think?

misalcedo (dismissed):

test/client_test.rb
misalcedo: Also, we should probably update the README with an example of adding retries. I don't think accepting a block on initialize is mentioned in the README at all (sorry about that), so an example of using it would be helpful documentation.

lib/elastomer_client/client.rb
```diff
@@ -283,11 +271,6 @@ def request(method, path, params)
         # wrap Faraday errors with appropriate ElastomerClient::Client error classes
         rescue Faraday::Error::ClientError => boom
           error = wrap_faraday_error(boom, method, path)
-          if error.retry? && RETRYABLE_METHODS.include?(method) && (retries += 1) <= request_max_retries
```
misalcedo, line 286: This is a bit odd, why do we only wrap client errors (i.e. 4xx status codes)? I would expect we would wrap everything and not just client errors.

See https://github.com/lostisland/faraday/blob/v0.17.6/lib/faraday/error.rb#L66-L68

misalcedo: Looking at [handle_errors](https://github.com/github/elastomer-client/blob/main/lib/elastomer_client/client.rb#L425-L445) that doesn't handle timeouts either. Also, there isn't anything in the [connection](https://github.com/github/elastomer-client/blob/main/lib/elastomer_client/client.rb#L140-L167) that raises errors, so Faraday shouldn't be raising client errors.

The more I look into this, the more confused I get as to how this works.

misalcedo: The only thing I can think of is that maybe [this meta-programming](https://github.com/github/elastomer-client/blob/main/lib/elastomer_client/client/errors.rb#L100-L109) is what enables retries. Though it shouldn't because it should skip anything that does not inherit from client error.

test/client_test.rb
kag728: Great idea on both ends! I'll add those today

test/client_test.rb
kag728: I added this back!

test/client_test.rb
kag728: Added!

misalcedo (approved):
LGTM

ndonewar (approved):

PR #287, opened 2023-12-05
Remove ES 8.6 support

ndonewar:
Removing ES 8.6 support as we've standardized on ES 8.7 and above. Verified version numbers in production, CI, and Codespaces.


test/version_support_test.rb
```diff
@@ -4,7 +4,7 @@
 
 describe ElastomerClient::VersionSupport do
   describe "supported versions" do
-    it "allows 5.0.0 to 8.6.0" do
+    it "allows 5.0.0 to 8.x" do
       five_series = ["5.0.0", "5.0.9", "5.2.0", "5.6.9", "5.6.100"]
       seven_series = ["7.0.0", "7.0.9", "7.2.0", "7.17.7", "7.17.100"]
       eight_series = ["8.0.0", "8.6.0"]
```
kag728, line 10: Can we update the eight series to include 8.7 instead of 8.6?

kag728: Yep, that works! I like it not tied to the specific version 👍

kag728 (approved):
:shipit:

kag728 (approved):

PR #288, opened 2023-12-05
Fix RuboCop warnings and default Ruby version

ndonewar:
This PR takes parts of [another PR](https://github.com/github/elastomer-client/pull/260) to focus on just the RuboCop fixes needed to get RuboCop in CI working.

- https://github.com/github/elastomer-client/commit/58dfe95e85e7f673161480d1f8bb1e713a5b6743 defaults the `.ruby-version` to the latest available
- https://github.com/github/elastomer-client/commit/ce36e5337aac4efc2c3097836bc0ee8fcd1545c2 is the largest change, but is also just the automatic fixes done by RubyCop (via `bundle exec rubocop -a`). 
- https://github.com/github/elastomer-client/commit/5731444c5a6a616ee57f4ce1f4286917ad7d42cb manually fixes the remaining 3 issues, which were all extraneous parentheses deletions.
- https://github.com/github/elastomer-client/pull/288/commits/59ef81fea76dbb662abf7d93755b5af5f4aab630 simplifies CI and RuboCop to 3.2 only


lib/elastomer_client/client.rb
```diff
@@ -86,12 +86,12 @@ def initialize(host: "localhost", port: 9200, url: nil,
     # fashion.
     def dup
       self.class.new \
-          url:              url,
-          read_timeout:     read_timeout,
-          open_timeout:     open_timeout,
+          url:,
+          read_timeout:,
+          open_timeout:,
```
tomthorogood, line 91: yay, i love seeing this shorthand ❤️

.github/workflows/main.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['3.0', '3.1', '3.2']
+        ruby-version: ['3.2']
         ES_VERSION: ['5.6.15', '8.7.0']
         include:
           - ES_VERSION: '5.6.15'
```
tomthorogood, line 17: Does this mean that we're no longer supporting 3.0 or 3.1 at all with elastomer client? Will that cause any customer impact, if users are depending upon that support?

tomthorogood (approved):
The changes make sense to me, I have one question about dependents outside of GitHub.

.github/workflows/main.yml
ndonewar: Yeah, since the library is only used internally, we're focusing support on the version we run.

PR #291, opened 2023-12-19
Move opaque id to the end of the middleware chain

richa-d:
This PR moves the call to the opaque ID middleware to the end of the Faraday middleware chain. Currently the opaque ID middleware is being triggered earlier in the middleware stack and is wrapping the invocation of the other middleware in the chain after it.
This change will ensure that we can use middleware like circuit breakers without causing issues for the opaque ID middleware. So any time we short circuit the remote request, the opaque ID middleware will not be called.

### DIFF
--- a/lib/elastomer_client/client.rb
+++ b/lib/elastomer_client/client.rb
@@ -134,7 +134,6 @@ module ElastomerClient
         # Request compressed responses from ES and decompress them
         conn.use(:gzip)
         conn.request(:encode_json)
-        conn.request(:opaque_id) if @opaque_id
         conn.request(:limit_size, max_request_size:) if max_request_size
         conn.request(:elastomer_compress, compression:) if compress_body
 
@@ -148,6 +147,8 @@ module ElastomerClient
         end
 
         @connection_block&.call(conn)
+        
+        conn.request(:opaque_id) if @opaque_id
 
         if @adapter.is_a?(Array)
           conn.adapter(*@adapter)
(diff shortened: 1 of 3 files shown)


kag728 (approved):
This is great! LGTM!

lib/elastomer_client/client.rb
```diff
@@ -149,6 +148,8 @@ def connection
 
         @connection_block&.call(conn)
 
+        conn.request(:opaque_id) if @opaque_id
```
misalcedo, line 151: The change looks good, but it is missing a new test. The test should cover creating a client with a middleware that always sets a fake response. The assertion would be that we get the fake response back (instead of an opaque id exception).

The simplest form of such a Faraday middleware is:

```ruby
module Elastomer
  module Middleware
    class MockResponse < Faraday::Middleware
      def initialize(app, &block)
        super(app)
        @response_block = block
      end

      def call(env)
        env.clear_body if env.needs_body?
     
        env.status = 200
        env.response_headers = ::Faraday::Utils::Headers.new
        env.response_headers["Fake"] = "yes"
        env.response = ::Faraday::Response.new

        @response_block&.call(env)

        env.response.finish(env) unless env.parallel?

        env.response
      end
    end
  end
end
```

The middleware above could be used like this:

```ruby
client = ElastomerClient::Client.new(**opts) do |connection|
  connection.request Elastomer::Middleware::MockResponse { |env| env.body = "{}" }
end
```

richa-d: Added a test. Thanks!

misalcedo (approved):
LGTM

PR #293, opened 2024-01-23
Change opaque ID middleware to not throw OpaqueIDError on 5xx responses

richa-d:
Currently, if ES returns a 500 response, we get an `OpaqueIDError` (and not the `ServerError`) since we only [check for an opaque ID mismatch](https://github.com/github/elastomer-client/blob/main/lib/elastomer_client/middleware/opaque_id.rb#L38) and don't handle the case when opaque ID is nil.
Here I added a check so that the `OpaqueIDError` will not be thrown if there is a 5xx response from Elasticsearch, and we will get a `ElastomerClient::Client::ServerError` instead.


kag728 (approved):
I really like this approach!

ndonewar (approved):
Thanks for the additional tests! 👍

PR #312, opened 2024-04-09
Update CI to 8.13.2

ndonewar:
Update CI to the same patch level we're planning to use throughout.

The .2 patch includes the fix for 8.13 that downgrades the JDK. More info:
https://www.elastic.co/blog/elastic-stack-8-13-2-released


kag728 (approved):
LGTM!

PR #309, opened 2024-04-05
Replace Faraday::Error::* usage

jasonkim:
- Replace `Faraday::Error::*` with `Faraday::*` error classes (following the [upgrade guide](https://github.com/lostisland/faraday/blob/main/UPGRADING.md#errors))
- Handle all `Faraday::Error` instead of `Faraday::Error::ClientError`
- Unlock faraday version to allow 1.0.0
- Bumped up the faraday and faraday middleware gems slightly

PR #316, opened 2024-06-05
Unlock faraday_middleware version

composerinteralia:
This is a followup to https://github.com/github/elastomer-client/pull/309

We can't upgrade faraday to 1.x without also upgrading faraday_middleware to 1.x, so the previous PR didn't get us all the way there.

Not much changes in faraday_middleware 1.x except for faraday 1.x support, so I expect elastomer-client to be fully compatible.

(We'll revisit this once more for faraday 2.0, which drops the faraday_middleware gem entirely, but that's a problem for another day.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 6.1.1 (2024-06-05)
+- Unlock faraday_middleware version to allow 1.x
+
 ## 6.1.0 (2024-04-09)
 - Replace `Faraday::Error::*` with `Faraday::*` error classes
 - Handle all `Faraday::Error` instead of `Faraday::Error::ClientError`
--- a/elastomer-client.gemspec
+++ b/elastomer-client.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "addressable", "~> 2.5"
   spec.add_dependency "faraday",     ">= 0.17"
-  spec.add_dependency "faraday_middleware", "~> 0.14"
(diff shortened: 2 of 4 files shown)


composerinteralia:
Looks like loosening the constraints caused the tests to run on new version and there's a failure, so I'll need to dig into that first.

composerinteralia:
Ah, OK, I think this is an "expected" failure. The adapter is now stored separately (it's been that way [for some time](https://github.com/lostisland/faraday/commit/8503baea4bf2bc821cf2c63125929c9fc7ef9d31#diff-14e9d694fa7c552f422f14b43d88d203f14a7a155c3c29e31a64c6125b8b7464L208), but we're newly using the new mechanism after upgrading.

Before:

```
#<Faraday::RackBuilder:0x00000001237fabb0 @handlers=[ElastomerClient::Middleware::ParseJson, FaradayMiddleware::Gzip, ElastomerClient::Middleware::EncodeJson, ElastomerClient::Middleware::LimitSize, Faraday::Adapter::Test]>
```

After:

```
#<Faraday::RackBuilder:0x00000001226b8e38 @adapter=Faraday::Adapter::Test, @handlers=[ElastomerClient::Middleware::ParseJson, FaradayMiddleware::Gzip, ElastomerClient::Middleware::EncodeJson, ElastomerClient::Middleware::LimitSize]>
```

So I'll update these tests to check that we're using the right adapter instead of checking that the adapter is in the list of handlers.

tomthorogood (approved):
🏎️ 🙇

PR #322, opened 2025-02-10
Add support for PUT _ccr/follow

jherns:
- Add support for following an index
- Update test framework to support CCR
  - will skip CCR tests if the replica isn't reachable
- Update readme

### DIFF
--- a/lib/elastomer_client/client/ccr.rb
+++ b/lib/elastomer_client/client/ccr.rb
@@ -32,7 +32,7 @@ module ElastomerClient
       #   ccr.follow("follower_index", { leader_index: "leader_index", remote_cluster: "leader" })
       #
       # See https://www.elastic.co/guide/en/elasticsearch/reference/current/ccr-put-follow.html
-      def follow(follower_index, body, params={})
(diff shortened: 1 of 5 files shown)


lib/elastomer_client/client/ccr.rb
```diff
@@ -0,0 +1,41 @@
...
+      #   :remote_cluster - String name of the remote cluster. Required.
+      #   :leader_index   - String name of the leader index. Required.
+      # params         - Hash of query parameters
+      #
+      # Examples
+      #
+      #   ccr.follow("follower_index", { leader_index: "leader_index", remote_cluster: "leader" })
+      #
+      # See https://www.elastic.co/guide/en/elasticsearch/reference/current/ccr-put-follow.html
+      def follow(follower_index, body, params = {})
```
dtaivpp, line 35: Should we at least check to make sure the following index was created here? 

` { "follow_index_created" : true }`

jherns: I think that is on the user (`elastomer`) to confirm the index was created.

thorrsson (approved):

PR #334, opened 2026-06-22
Support Faraday 2 alongside 1.x

Tonkpils:
## What

Makes elastomer-client work on both Faraday 1.x and 2.x

## Why

elastomer-client's faraday constraint was already open (`>= 0.17`), but it depended on the deprecated `faraday_middleware`, which pins faraday to `~> 1.0`. The only thing actually used from it was the `:gzip` middleware.

## Changes

- Swap `faraday_middleware` for `faraday-gzip` (works on both 1.x and 2.x). It registers `:gzip` on `Faraday::Request`, so `conn.use(:gzip)` becomes `conn.request(:gzip)`.
- Faraday 2 removed `Connection#token_auth`/`#basic_auth`. The `:authorization` middleware isn't a clean swap (its API and header output differ across majors), so set the `Authorization` header directly. Verified byte-identical to Faraday 1.x's `token_auth`/`basic_auth` output (`Token token="..."` / `Basic <base64>`).
- gemspec: `faraday >= 1.0, < 3`, drop `faraday_middleware`, add `faraday-gzip`. Bump to 6.3.0.
- All 5 custom middleware already use Faraday 2 patterns (`Faraday::Middleware` + `@app.call`/`on_complete`), no changes needed.

## Test changes (necessitated by the above)

- Auth tests spied on the removed `token_auth`/`basic_auth` methods — now assert the resulting `Authorization` header (no live server needed). The `spy` dev dep is no longer used.
- The block-config test referenced the gone `FaradayMiddleware::Instrumentation` — swapped for elastomer's own `:opaque_id`.
- The retry test looked up the relocated retry class via the middleware registry.

## CI

Added a Faraday 1.x / 2.x gemfile matrix so both majors are tested against the same ES versions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,17 +12,20 @@ jobs:
     strategy:
       matrix:
         ruby-version: ['3.2']
+        gemfile: ['faraday_1', 'faraday_2']
         ES_VERSION: ['8.18.0']
         include:
           - ES_VERSION: '8.18.0'
             ES_DOWNLOAD_URL: >-
               https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-8.18.0-linux-x86_64.tar.gz
+    env:
+      BUNDLE_GEMFILE: ${{ github.workspace }}/gemfiles/${{ matrix.gemfile }}.gemfile
     steps:
+      - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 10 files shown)


tomthorogood (approved):
@Tonkpils Thanks for doing this; we will also need to update github/github to use the new version of this client once it is released. I can take care of that, just wanted to let you know it won't flow in automatically on merge. ❤️

